PR #6, opened 2015-12-18
Pretty Ticks

Frencil:
This branch implements R's `pretty` function to generate tick values in multiples of 1, 2, and 5, scaled to the order of magnitude of the range. It then applies pretty to the default axis rendering for x, y1, and y2, while also scaling y1 and y2 domains to the generated pretty tick values.

Still needs unit tests, but it looks to work about as expected for all the top hits.

[R's pretty function source code](https://github.com/wch/r-source/blob/b156e3a711967f58131e23c1b1dc1ea90e2f0c43/src/appl/pretty.c)


MrFlick:
Look good to me. Let's try it out and see how it looks.

PR #7, opened 2015-12-21
Datainit

MrFlick:
These changes update the interaction for specifying API data endpoints that are consumed by the data. This introduces a new LocusZoom.DataSources() object that allows for adding namespace/datasource pairs either though chainable .addSource() calls, or through an object literal syntax via .setSource() or via a JSON string.

This also updates the initialization methods populate() and populateAll(). These should be the primary means of create LZ objects. They have more broad support for D3 selectors and have separate paramters for data sources, plot layout, and initial state (though the later two are not yet implemented).

### DIFF
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -4,7 +4,58 @@
 
 "use strict";
 
+/* A named collection of data sources used to draw a plot*/
+LocusZoom.DataSources = function() {
+    this.sources = {};
+};
+
+LocusZoom.DataSources.prototype.addSource = function(ns, x) {
+    function findKnownSource(x) {
+        if (!LocusZoom.KnownDataSources) {return null;}
+        for(var i=0; i<LocusZoom.KnownDataSources.length; i++) {
+            if (!LocusZoom.KnownDataSources[i].sourcename) {
+                throw("KnownDataSource at position " + i + "does not have a 'sourcename'");
+            }
+            if (LocusZoom.KnownDataSources[i].sourcename == x) {
+                return LocusZoom.KnownDataSources[i];
+            }
+        }
+        return null;
+    };
+    if (Array.isArray(x)) {
+        var dsclass = findKnownSource(x[0]);
+        if (dsclass) {
+            this.sources[ns] = new dsclass(x[1]);
+        } else {
+            throw("Unable to resolve " + x[0] + " data source");
+        }
+    } else {
+        this.sources[ns] = x;
+    }
+    return this;
+};
+
+LocusZoom.DataSources.prototype.getSource = function(ns) {
+    return this.sources[ns];
+};
+
+LocusZoom.DataSources.prototype.setSources = function(x) {
+    if (typeof x === "string" || x instanceof String) {
+        x = JSON.parse(x);
+    }
+    var ds = this;
+    Object.keys(x).forEach(function(ns) {
+        ds.addSource(ns, x[ns]);
+    });
+    return ds;
+};
+
+LocusZoom.DataSources.prototype.keys = function() {
+    return Object.keys(this.sources);
+}
+
 LocusZoom.Data = LocusZoom.Data ||  {};
+
 
 LocusZoom.Data.Requester = function(sources) {
 
@@ -81,6 +132,7 @@ LocusZoom.Data.AssociationSource = function(url) {
         };
     };
 };
+LocusZoom.Data.AssociationSource.sourcename = "AssocationLZ";
 
 LocusZoom.Data.LDSource = function(url) {
     this.url = url;
@@ -143,6 +195,7 @@ LocusZoom.Data.LDSource = function(url) {
         };
     };
 };
+LocusZoom.Data.LDSource.sourcename = "LDLZ";
 
 LocusZoom.Data.GeneSource = function(url) {
     this.url = url;
@@ -161,6 +214,7 @@ LocusZoom.Data.GeneSource = function(url) {
         };
     };
 };
+LocusZoom.Data.GeneSource.sourcename = "GeneLZ";
 
 LocusZoom.createResolvedPromise = function() {
     var response = Q.defer();
@@ -168,6 +222,10 @@ LocusZoom.createResolvedPromise = function() {
     return response.promise;
 };
 
+LocusZoom.KnownDataSources = [
+    LocusZoom.Data.AssociationSource,
+    LocusZoom.Data.LDSource,
+    LocusZoom.Data.GeneSource];
 LocusZoom.DefaultDataSources = {};
 
 
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -13,10 +13,10 @@ LocusZoom.DataSources.prototype.addSource = function(ns, x) {
     function findKnownSource(x) {
         if (!LocusZoom.KnownDataSources) {return null;}
         for(var i=0; i<LocusZoom.KnownDataSources.length; i++) {
-            if (!LocusZoom.KnownDataSources[i].sourcename) {
-                throw("KnownDataSource at position " + i + "does not have a 'sourcename'");
+            if (!LocusZoom.KnownDataSources[i].SOURCE_NAME) {
+                throw("KnownDataSource at position " + i + " does not have a 'SOURCE_NAME' static property");
             }
-            if (LocusZoom.KnownDataSources[i].sourcename == x) {
+            if (LocusZoom.KnownDataSources[i].SOURCE_NAME == x) {
                 return LocusZoom.KnownDataSources[i];
             }
         }
@@ -64,11 +64,11 @@ LocusZoom.Data.Requester = function(sources) {
         fields.forEach(function(field) {
             var parts = field.split(/\:(.*)/);
             if (parts.length==1) {
-                if (typeof requests["position"] == "undefined") {
-                    requests.position = {names:[], fields:[]};
+                if (typeof requests["base"] == "undefined") {
+                    requests.base = {names:[], fields:[]};
                 }
-                requests.position.names.push(field);
-                requests.position.fields.push(field);
+                requests.base.names.push(field);
+                requests.base.fields.push(field);
             } else {
                 if (typeof requests[parts[0]] =="undefined") {
                     requests[parts[0]] = {names:[], fields:[]};
@@ -83,10 +83,10 @@ LocusZoom.Data.Requester = function(sources) {
     this.getData = function(state, fields) {
         var requests = split_requests(fields);
         var promises = Object.keys(requests).map(function(key) {
-            if (!sources[key]) {
+            if (!sources.getSource(key)) {
                 throw("Datasource for namespace " + key + " not found");
             }
-            return sources[key].getData(state, requests[key].fields, requests[key].names);
+            return sources.getSource(key).getData(state, requests[key].fields, requests[key].names);
         });
         //assume the fields are requested in dependent order
         //TODO: better manage dependencies
@@ -132,7 +132,7 @@ LocusZoom.Data.AssociationSource = function(url) {
         };
     };
 };
-LocusZoom.Data.AssociationSource.sourcename = "AssocationLZ";
+LocusZoom.Data.AssociationSource.SOURCE_NAME = "AssocationLZ";
 
 LocusZoom.Data.LDSource = function(url) {
     this.url = url;
@@ -195,7 +195,7 @@ LocusZoom.Data.LDSource = function(url) {
         };
     };
 };
-LocusZoom.Data.LDSource.sourcename = "LDLZ";
+LocusZoom.Data.LDSource.SOURCE_NAME = "LDLZ";
 
 LocusZoom.Data.GeneSource = function(url) {
     this.url = url;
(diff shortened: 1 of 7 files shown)


assets/js/app/LocusZoom.js
```diff
@@ -151,5 +210,13 @@
         return response.promise;
     };
 
+    exports.isString = function(x) {
```
Frencil: I don't think this method is necessary, and it's tough to argue this is a thing that the LocusZoom singleton should "do" (in other words the LocusZoom singleton does a lot of things like populate divs with instances or parse position queries, but type-checking is something that core JS should be doing).

In JavaScript the only time you'd need to check the value with `instanceof` is if the value is an object-wrapped string (instantiated with the `x = new String("foo");` syntax, which we should really never do). As such a simple `typeof x === "string"` is sufficient wherever we need to check for stringyness.

assets/js/app/Data.js
```diff
@@ -4,8 +4,60 @@
...
+    }
+    return this;
+};
+
+LocusZoom.DataSources.prototype.getSource = function(ns) {
+    return this.sources[ns];
+};
+
+LocusZoom.DataSources.prototype.setSources = function(x) {
+    if (LocusZoom.isString(x)) {
```
Frencil: [see note](https://github.com/statgen/locuszoom/pull/7/files#diff-8c4275047bc760f446e4203cd9f5e5fdR213)

assets/js/app/Data.js
```diff
@@ -47,9 +99,19 @@ LocusZoom.Data.Requester = function(sources) {
     };
 };
 
-LocusZoom.Data.AssociationSource = function(url) {
-    this.url = url;
-
+LocusZoom.Data.AssociationSource = function(init) {
+    if (LocusZoom.isString(init)) {
```
Frencil: [see note](https://github.com/statgen/locuszoom/pull/7/files#diff-8c4275047bc760f446e4203cd9f5e5fdR213)

assets/js/app/Data.js
```diff
@@ -81,9 +144,19 @@ LocusZoom.Data.AssociationSource = function(url) {
         };
     };
 };
+LocusZoom.Data.AssociationSource.SOURCE_NAME = "AssociationLZ";
 
-LocusZoom.Data.LDSource = function(url) {
-    this.url = url;
+LocusZoom.Data.LDSource = function(init) {
+    if (LocusZoom.isString(init)) {
```
Frencil: [see note](https://github.com/statgen/locuszoom/pull/7/files#diff-8c4275047bc760f446e4203cd9f5e5fdR213)

assets/js/app/Data.js
```diff
@@ -143,13 +217,24 @@ LocusZoom.Data.LDSource = function(url) {
         };
     };
 };
+LocusZoom.Data.LDSource.SOURCE_NAME = "LDLZ";
 
-LocusZoom.Data.GeneSource = function(url) {
-    this.url = url;
+LocusZoom.Data.GeneSource = function(init) {
+    if (LocusZoom.isString(init)) {
```
Frencil: [see note](https://github.com/statgen/locuszoom/pull/7/files#diff-8c4275047bc760f446e4203cd9f5e5fdR213)

assets/js/app/Data.js
```diff
@@ -4,8 +4,60 @@
 
 "use strict";
 
+/* A named collection of data sources used to draw a plot*/
+LocusZoom.DataSources = function() {
```
Frencil: Would it make more sense to nest this inside the Data object, as in `LocusZoom.Data.Sources`? This would keep a more consistent *file* architecture where each app file completely defines one and only one class in the singleton's namespace. Otherwise this is the only file (so far) that defines two: `LocusZoom.Data` and `LocusZoom.DataSources`.

assets/js/app/LocusZoom.js
MrFlick: I've been treating "LocusZoom" more like a namespace than a singleton. I don't want to pollute the global JS environment with functions that might conflict. For example, Angular has it's own isString. Ideally it's just be a private function not exposed but we didn't make that easy with our file layout. It would actually be easier if we only had one JS file which is something we might want to revisit. I think isString() is much easier to read and understand the intention of rather than `typeof x` and it matches better with `Array.isArray()`. We could add it to the String prototype, but again i'd like to limit changes to behave to things only within the LocusZoom namespace.

assets/js/app/Data.js
MrFlick: I've really only been treating "Data" as a namespace to try to not pollute the main LocusZoom namespace. There's not really a need for it and it should probably disappear. Rather than LocusZoom.Data.GeneSource. It should be LocusZoom.GeneDataSource or something.

assets/js/app/Data.js
MrFlick: These say "see note" but i can't find the note.

assets/js/app/LocusZoom.js
```diff
@@ -15,35 +15,48 @@
     // Create a new instance by instance class and attach it to a div by ID
     // NOTE: if no InstanceClass is passed then the instance will use the Intance base class.
     //       The DefaultInstance class must be passed explicitly just as any other class that extends Instance.
-    exports.addInstanceToDivById = function(InstanceClass, id){
+    exports.addInstanceToDivById = function(id, datasource, layout, state){
```
MrFlick: As you trying to say that "addInstanceToDivById" should stay exported? I think this function should not be exposed by the API.

assets/js/app/LocusZoom.js
Frencil: I'm not suggesting creating a global `isString` function, I'm suggesting using the basic `typeof x === "string"` syntax. One could argue it's not quite as clean as wrapping in an `isString` method of some kind, but it's standard JS string type checking and therefore something that javascript developers should know and be familiar with.

If this general syntax is a big issue for you then by all means we can keep the method. At the very least it can be less verbose:

```javascript
exports.isString = function(x) {
  return (typeof x === "string");
}
```

Since any time you're evaluating a conditional and returning true or false you can just as easily return the evaluated conditional.

But in general I favor caution when wanting to define functions in our narrowly-scoped library purely for making javascript's general syntax prettier. Like all languages it has its warts and we may find ourselves on a slippery slope defining more syntax helpers if we're not careful.

assets/js/app/Data.js
Frencil: Clicking the [see note](https://github.com/statgen/locuszoom/pull/7/files#diff-8c4275047bc760f446e4203cd9f5e5fdR213) link should take you to the line in this pull req's diff where the `isString` function is defined (where our discussion of the function lives). I just threw this on all the instances of that functions usage that I could find.

assets/js/app/LocusZoom.js
Frencil: I wasn't really try to make any architectural statement; I was just trying to fix the broken build! See [this line](https://github.com/statgen/locuszoom/pull/7/files#diff-29448d642d7578e29c0da8c8b0382e91R62) in the test suite.

In order for that test to work that method needs to be exposed. Now tests should always evolve as the architecture evolves and not the other way around, so if it's "correct" to obscure that method then this test needs to change. Either way the build has to pass in order to merge.

As to the discussion of "what to export" I hadn't made that leap and I don't think we need to be particularly concerned with it right now either way. The only reason I refactored the base `LocusZoom.js` into a JS module with the exports syntax is so that mocha could pull it in as a module and we could have some automated tests. It worked just fine before without being a module at all. Now that it's built as a module we *can* get specific about what we expose, but I'd argue it's more of a distraction at this point. Just exposing every method is way simpler (yay KISS principle) and makes it easier to build unit tests for every method we build. Pulling it from the documentation and favoring other methods (as you've done in this branch) is a far more effective approach for our tiny user base, and if we actually have users digging through the source and plugging in undocumented functions in ways we don't expect them to then that's not our problem.

assets/js/app/LocusZoom.js
```diff
@@ -52,6 +65,52 @@
...
+                mult = 1e3; //K
+            }
+            val = val.replace(suffixre,"");
+        }
+        val = Number(val) * mult;
+        return val;
+    }
+
+    // Parse region queries that look like
+    // chr:start-top
```
Frencil: start-stop?

demo.html
```diff
@@ -11,11 +11,11 @@
...
 
   <script type="text/javascript">
-    LocusZoom.DefaultDataSources = {
-        position: new LocusZoom.Data.AssociationSource("http://portaldev.sph.umich.edu/api_internal_dev/v1/single/"),
-        ld: new LocusZoom.Data.LDSource("http://portaldev.sph.umich.edu/api_internal_dev/v1/pair/LD/"),
-        gene: new LocusZoom.Data.GeneSource("http://portaldev.sph.umich.edu/api_internal_dev/v1/annotation/genes/")
-    };
+    var apiBase = "http://portaldev.sph.umich.edu/api_internal_dev/v1/";
+    var ds = new LocusZoom.DataSources();
+	ds.addSource("base", ["AssociationLZ", {url:apiBase + "single/", params: {analysis: 3}}]);
```
Frencil: So unlike "LD" and "Gene" the positions/base/associations datasource/datalayer/panel seem to have a lot of names. Well, not that many. And naming things is one of the [two hard things](http://martinfowler.com/bliki/TwoHardThings.html). But does it feel like there's any convergence on a common name here? Quite probably all the names in play are totally legit for how they're used. Suffice it to say when I named that one panel the `PositionsPanel` I wasn't sure if it was the best name, and I'm still not. So if there is a more globally-applicable name for the main stuff rising above the rest I'd be happy to help apply it more globally.

assets/js/app/LocusZoom.js
Frencil: Edit: just remembered that **end** is what we've settled on as opposed to **stop**

assets/js/app/Data.js
```diff
@@ -47,9 +99,19 @@ LocusZoom.Data.Requester = function(sources) {
...
+        this.params = {};
+    } else {
+        this.url = init.url;
+        this.params = init.params || {};
+    }
+    if (!this.url) {
+        throw("AssociationSource not initialized with required URL");
+    }
+    
+    var that=this;
```
Frencil: If you're not a huge fan of cloning `this` into another var (`that`) to have as a reference in a scoped function then `.bind()` can help. This of course only works when the scoped function isn't already referring to itself with `this`, which is the case for `this.getData()`. The syntax would go something like:

```javascript
this.getData = function(state, fields, outnames) {
    // stuff, including references to "this"
}).bind(this);
```

It's arbitrary which we use. My reasons for favoring `bind()` are that `this` is a known special keyword and intuitively when working with a function's logic I expect `this` to refer to an object or thing that's a major part of my context. If that same context comes from another variable it works the same, but I need to keep that other variable's name in mind, which can just be a bit less intuitive.

assets/js/app/Data.js
```diff
@@ -112,6 +185,7 @@ LocusZoom.Data.LDSource = function(url) {
         }
     };
 
+    var that = this;
```
Frencil: [See other note regarding this/that/bind](https://github.com/statgen/locuszoom/pull/7/files#r48358203)

assets/js/app/Data.js
```diff
@@ -143,13 +217,24 @@ LocusZoom.Data.LDSource = function(url) {
...
+        this.params = {};
+    } else {
+        this.url = init.url;
+        this.params = init.params || {};
+    }
+    if (!this.url) {
+        throw("GeneSource not initialized with required URL");
+    }
 
+    var that = this;
```
Frencil: [See other note regarding this/that/bind](https://github.com/statgen/locuszoom/pull/7/files#r48358203)

assets/js/app/LocusZoom.js
MrFlick: I thought i commented the addInstanceToDivById test out of the test file in my code. When I ran it all tests passed. I think this really should be a module but being a module requires everything to be in the same file. The fact that we started with different files was probably the wrong choice and has made many things more difficult.

assets/js/app/LocusZoom.js
MrFlick: I've found it very helpful to have a set of consistent isString, isFunction, isArray methods. I think your suggested edit is good. It would be even better to keep those as private functions and not expose them outside the module.

demo.html
MrFlick: I think it will be more clear how everything ties together when we extract out the layout object since that will directory connect to the datasource names. Right now I didn't change too much so as to not break too many things. The other big thing we need to rename is "Instance." That sticks out like a sore thumb to me when we talk about an "instance of Instance". Right now i'm leaning more toward "Plot". It's a bit more generic than i like but i think it's better than abusing "instance." Naming is hard.

assets/js/app/Data.js
MrFlick: Good suggestion. I had not used that syntax before. Does it work well with all the promise stuff?

assets/js/app/LocusZoom.js
MrFlick: Ah, i had removed the test but it got re-added when i merged with master at one point.

assets/js/app/Data.js
Frencil: Other libraries shouldn't be affected by this because each time you bind variables to a function's scope you do it as you define that function and it only applies to the scope of that function. In other wordsm binding `this` on a function you define that invokes methods on the Q library will have no effect on when those Q methods use the `this` keyword because your binding only applies to the scope of the function you created. So this should play nice with everything. =)

MrFlick:
I've made most of the coding changes that were addressed. I've held of on any renaming or reorganization issues for now.

PR #11, opened 2016-01-05
Object Oriented Data Sources

MrFlick:
I tried to make the Data.Source classes a bit more friendly and easier to customize. I introduced a base class for them to inherit from. I separated the main functions of building the URL for the request and parsing the response to methods on the prototype that can be replaced or re-used for other data sources. This update also allows the DataSources collection to be serialized to JSON (`JSON.stringify(ds)`) and later re-inflated (`ds.setSources()`).


Frencil:
Looks good!

PR #12, opened 2016-01-14
Curtain

Frencil:
## Curtain

Ideally, when things go wrong, we need a way of showing that in the plugin itself (logging to the console is not a graceful failure if the console is not open). This branch introduces a new top-level group to an instance's SVG classed `lz-curtain`. It contains an 85% opaque grey rect overlay to obscure content and interactive elements regardless of the state they're in and pipes a customizable message on top.

The curtain is a top-level object of an instance object and it has the following properties:

* `instance.curtain.svg` - selector for the `<g>` element containing the curtain
* `instance.curtain.drop(message)` - takes a string message as an argument, makes the curtain appear with said message.
* `instance.curtain.raise()` - removes the curtain overlay.

As we add interactive elements we'll want to establish a pattern of wrapping everything in try/catch such that when something goes wrong we can gracefully obscure the instance and display context to the end user.


Frencil:
NOTE: This branch depends on the fixed testing environment in Pull #13. Don't merge yet!

Frencil:
With #13 merged this branch has now pulled in the updates to the testing environment and is filled out with working tests.

Also per discussion the curtain no longer appears during the initialization of a given plugin instance.

**Ready for Review and Merge.**

Frencil:
Update: I just added curtains on panels that behave the exact same as curtains on the overall instance. In practice these may end up being more commonly seen - for instance, the panel `reMap()` method is now wrapped in a try/catch to pipe errors with the data layers or rendering to the curtain.

As such, moving forward, we can rely on the easy development pattern of `throw()` when encountering a problem that should bubble up throughout the data getting/processing and user interaction elements of the plugin. We can be increasingly confident that those errors should be gracefully caught and displayed to the end user.

MrFlick:
So it this ready for merging now, or will there be more updates?

Frencil:
No more updates coming in on this one. Note, though that there are slight conflicts between #12 and #14 so whichever you merge first I'll probably need to do a quick resolution on the other. That said, both are complete and ready to go whenever.

PR #14, opened 2016-01-17
Purge instance storage

Frencil:
This branch removes the `_instances` object from the LocusZoom singleton. the reasoning here is that much like our main two dependencies (d3 and Q) this library extends an API for creating and working with a complex object but how that object is stored in memory is up to the developer using the plugin to implement a specific use case.

This is a breaking change, assuming previous LocusZoom plugin implementations relied on plugin instance storage.


Frencil:
Just cleaned up the one tiny conflict this had with #12. Should be good to review and merge now.

Issue #19, opened 2016-01-18
Resizable instances


Frencil:
**Depends on #15**

Once a recursive redraw methods pattern is in place (#15) we can look at making instances themselves resizable. The ideal behavior here would be a standard-looking "resize" handle appearing in the lower right corner with click/drag behavior that resizes the entire SVG element and all panels/data therein proportionally.

Note that the [CSS3 resize](http://www.w3schools.com/cssref/css3_pr_resize.asp) property won't help us here. This works well for HTML elements but is not implemented for SVG. We'll need to implement a UI layer for showing the resize handle we draw ourselves, and use mouse event listeners to implement the resize behavior.

## Estimate

This feature will probably take about a week of development to build and test.

Issue #15, opened 2016-01-18
Recursive transitional re-draw methods


Frencil:
For two major features (dynamic resizing of instances/panels and panning/zooming) we need recursive tiers of transitional re-draw methods.

## The problem

Essentially right now all svg elements are rendered once. There are API endpoints to change the sizes of elements or to map to a new section of data but in all cases existing elements must be dropped and drawn again from scratch.

## The solution

We need similar methods on all nested objects (instance, panel, data layer) that assume all svg elements already exist and, using updated `state` data, define proper d3 transitions to smoothly animate the resizing/positioning of elements without destroying/recreating them.

## Estimate

This feature will probably take a dedicated week of development to build and test.

PR #24, opened 2016-01-21
Relative sizing

Frencil:
# Relative Sizing

## Feature Overview

1. Relative sizing for all panel and instance level elements
2. Hard distinction between initializing and rendering SVG elements + cascading render calls
3. Proportional sizing for panels with default behavior of equal sizing
4. Introduction of a UI layer at the instance level
5. First instance-level UI element: drag handle in lower-right corner to resize instance

## Issues

Addresses #15 and #19.

### 1. Relative sizing for all panel and instance level elements

Before this branch many panel-level elements were created with hard numbers for size and positioning. After this branch all elements are positioned and sized using relative values from the `view` object on the instance or the panel.

Example: in `LocusZoom.Panel.prototype.render` the translate values for all groups containing axis and tick lines, as well as the x/y positioning of labels, is all done relative to values found in the `view` object such as the width/height of the panel of the panel's clip area and/or the panel's margins.

### 2. Hard distinction between initializing and rendering SVG elements + cascading render calls

**Initialize** and **Render**, conceptually, should do two very different things:

* **Initialize** should create new elements in the SVG with IDs, classes, and other attributes that tend not to change, as well as store those SVG selectors for quick access by other methods (like render).
* **Render** should use current state/view/whatever values to generate and apply position values (x, y, transform/translate, etc) and any other values that will tend to change with user input.

The above is a general pattern that should appear at each level, more or less. In this branch we apply it to instances and panels. Another way of describing this is initialize should be something we only do *once*, whereas *render* is something we do *as often as we need to*, usually when associated values change (like changing the dimensions on the entire instance).

### 3. Proportional sizing for panels with default behavior of equal sizing

In this branch the `view` object on panels now has `proportional_width` and `proportional_height`. These are values between 0 and 1 to be treated as percentages and default to 1 and 1 (e.g. by default a panel on its own will take up the entire instance).

There are hooks in place in this branch to generate the dimensions and offset of a given panel in pixels using these proportional values along with instance-level width/height values.

There is also a new instance method: `Instance::stackPanels()` which automatically sets the proportional heights, offsets, and dimensions of all panels equally in the order they were added. Note that, in the current demo, this *exactly* emulates what was already being done, just with a different logical approach. Methods to change relative panel sizes/offsets after the fact should still work the same.

We discussed more sophisticated logic for panel sizing, in particular how the Genes panel needs a more "quantized" height based on its data at the moment and shouldn't be "stretchy". These are good ideas but arguably out of scope for this branch, so it's best if we keep the rework of this logic to a minimum for now. Think of this as merely a first iteration.

### 4. Introduction of a UI layer at the instance level

There's a new top-level group that sits between all panels and the instance curtain called `ui`. It has an associated object on an instance (`instance.ui`) which houses svg selectors and methods.

The UI layer in this branch is not shown be default, but upon initialization adds mouseover/mouseout event handlers to make it appear when the mouse is over the SVG. It too has a separate `render` function for actually positioning elements on the UI layer, as repositioning is required whenever the instance's dimensions change.

### 5. First instance-level UI element: drag handle in lower-right corner to resize instance

Lastly, the UI layer creates one functional UI element: the drag handle. This appears in the lower-right corner and allows a user to change the dimensions of the entire instance. Panels will scale proportionally, keeping their 50/50 split of the available height, setting off cascading render calls throughout the panels (and to the ui layer itself). This is an excellent example of why there's a hard distinction between initialize/render methods, as resizing with the handle can trigger literally dozens of render calls in under a second.

## Version Bump: 0.2

This branch has a lot of changes to rendering and the introduction of some functional UI, so it warrants a version bump to **0.2**.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ node_js:
   - "4.2"
 
 before_script:
-  - npm install gulp gulp-util gulp-watch gulp-concat gulp-uglify gulp-mocha gulp-sass gulp-wrap yargs
-  - npm install mocha should jsdom mocha-jsdom
+  - npm install gulp@3.9.0 gulp-util@3.0.7 gulp-watch4.3.5 gulp-concat@2.6.0 gulp-uglify@1.5.1 gulp-mocha@2.2.0 gulp-sass@2.1.1 gulp-wrap@0.11.0 yargs@3.32.0
+  - npm install mocha@2.3.4 should8.0.1 jsdom@7.2.1 mocha-jsdom@1.0.0
 
 script:
     gulp
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ node_js:
   - "4.2"
 
 before_script:
-  - npm install gulp gulp-util gulp-watch gulp-concat gulp-uglify gulp-mocha gulp-sass gulp-wrap
+  - npm install gulp gulp-util gulp-watch gulp-concat gulp-uglify gulp-mocha gulp-sass gulp-wrap yargs
   - npm install mocha should jsdom mocha-jsdom
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ node_js:
   - "4.2"
 
 before_script:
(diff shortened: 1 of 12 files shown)


Frencil:
NOTE: For some reason today TravisCI is failing this build because `nvm install 4.2` fails and it reverts to installing node from source, ending up with a *really* old version (0.10.36). The vast majority of the node modules our build depends on then simply fail.

I've double-checked and this is not an issue with our code. Why TravisCI is failing to download node 4.2 only on branches is something only the devs over at TravisCI can figure out. They'll likely resolve it in time, but suffice it to say that as of this comment's writing the build is, in fact, passing.

Issue #22, opened 2016-01-18
Export / PDF Generation


Frencil:
A potential use case for LocusZoom in the future is means of exporting a LocusZoom instance to a graphic or PDF suitable for inclusion in a paper.

This feature is arguably out of scope of the plugin itself and will require a companion repository for server-side installation.

## Estimate

This feature will probably take at least a month of development to build and test because it branches out into a pretty different technology stack from the plugin itself.

Frencil:
**Update:** A proof of concept of this is now on the [export](https://github.com/statgen/locuszoom/tree/export) branch. It turns out this does *not* need to be a separate server-side module.

As demonstrated in [ExAC](http://exac.broadinstitute.org/) one stream an SVG element into a Base64-encoded string and then into a link for download as a data stream entirely on the client side.

This task will still take some work to be a fully-functional feature in the plugin itself, but it can definitely be done and will be simpler in scope.

Frencil:
**Update:** This is now addressed in #27 but is blocked by #30. That PR establishes a new syntax for serializable layouts which will allow for more intuitive optional inclusion of the controls element (and developer customization thereof) by refactoring this feature as an optional "HTML module".

The same pattern is being proposed for #21, which also has #30 as a blocker.

Frencil:
Released in v0.3.9.

PR #27, opened 2016-01-25
Export / Download SVG

Frencil:
As documented in Issue #22 users will need a way to "export" a LocusZoom plugin at a certain state to a graphic that can be used in papers.

This branch implements that feature entirely on the client side by cloning the SVG element, stripping away browser-interaction-specific elements, embedding the global LocusZoom CSS, and streaming into the `href` attribute of a link for download as an SVG file.

## Update - 2016-02-10

**Problems**

The initial implementation involved regenerating the base64 encoded SVG string each time the SVG changed (new data, resizing instance, etc.) While this didn't create a performance hit per se it certainly had the potential to do so as the plugin continues to mature. It also inserted infrequently necessary logic in frequently executed routines. There was also debate as to whether such a download button should even be a part of the plugin or is better suited as a partially-built solution with no graphical front end such that creating a working download SVG method is left to the implementor.

**Solutions**

First off, the base64 encoded SVG string is now generated with a promise that fires when the mouse moves over the download button. The button is disabled while the string is being generated to ensure broken downloads don't crop up as the base64 string generation time gets longer.

Secondly, I made the `controls` element an optional thing tied to a boolean on the instance called `show_controls`. This boolean defaults to `false` on the abstract Instance class and `true` on the DefaultInstance class. Where previously we had hooks in place to regenerate the base64 string we now just flip a boolean on the controls object that the SVG has changed, and *only* if `show_controls` is true on the Instance.

**General thoughts on inclusion/exclusion of features**

While I agree that making the controls element a mainstay for all LocusZoom instances in all circumstances *perhaps* adds unnecessary functionality in *some* use cases, I disagree that it's something to remove and leave up to an implementing developer to handle.

Fundamentally LocusZoom is a javascript plugin that you point at one or more `<div>` elements and it fills them with stuff. That stuff should not be restricted to only the core `<svg>` element because, as already has been discussed, SVG can be a poor medium for presenting text-heavy / tabular information or form elements. As long as everything is kept within the parent `<div>` then that is LocusZoom's responsibility. This alone should justify the creation of HTML elements adjacent to the SVG such as the controls element introduced in this branch and the pop over / info box elements forthcoming.

As for the download button itself it's been vociferously identified by many LocusZoom consumers already that this is a desirable feature. Making a pure javascript/CSS implementation of this feature an optional addition addresses the needs of users now who want it *and* hypothetical future users who *may* not want it. It's entirely possible that down the line a LocusZoom consumer working in an Angular environment wants part of the controls element but not all, or some other scenario we can imagine but are not directly facing today. We'll never predict or be able to build to [every possible use case](http://www.xkcd.com/1172/), and thus it's important to be pragmatic in our decision making. This optional download button belongs in the plugin because it implements a fully-client side solution consistent with the architecture of the plugin to a real problem expressed by the current community of LocusZoom consumers.


Frencil:
Putting this branch on hold. #30 - new layout syntax will allow for easier creation of HTML modules like this controls element, info box elements, etc. with more than just a boolean for enabling/disabling (e.g. toggling/customizing individual elements contained within). That should go first so this can be implemented in a more configurable way.

Frencil:
We're at a good point to bring this branch back into the fold with a layout parameter that defaults to not showing the controls object. It may also be worth rethinking the general controls approach, but still finding a way to solve the problem of users that want to export/download what they see as a static image.

Building a demo page suggested by #46 will provide a means to just develop the library side of this branch without also adding more demo pages in the scope of this branch, so that's now a dependency.

Frencil:
Closing this branch. It diverged enough from master and had a small enough footprint to manually spin up a new branch to address this need.

PR #29, opened 2016-02-19
Data Source Field Transformations

MrFlick:
This update allows transformations to be applied to fields returned from the API.

For example, on the demo page if you run

    var q = new LocusZoom.Data.Requester(ds);
    var state = {chr:10,start:114550452,end:114555452};
    var dumpbody = function(x) {console.log(x.body)};

You can then either call

    q.getData(state,["position","pvalue"]).then(dumpbody).done()

or

    q.getData(state,["position","pvalue|neglog10"]).then(dumpbody).done()

Transformations are indicated by pipes (`|`) followed by a function name. Functions can be added to `LocusZoom.KnowTransformations` which is an object that maps strings to functions.


assets/js/app/Data.js
```diff
@@ -281,3 +288,43 @@ LocusZoom.KnownDataSources = [
...
+                }
+                return val;
+            };
+        }
+    } 
+    return null;    
+};
+
+// This of known transformations (thus it can be extended)
+LocusZoom.Data.KnownTransformations = {
```
Frencil: First impression is that this is a nice intuitive pattern for such functions. "Known" may not be necessary in the name since they're just Transformations and if one isn't found/known we'll be [throwing an exception anyway](https://github.com/statgen/locuszoom/pull/29/files#diff-a6139e2504656ffc08c847116c79ce29R302).

We might also want to consider adding a `LocusZoom.Data.addKnownTransformation` (or just `addTransformation`) method to let users define their own too. I can definitely see how Transformations in general would be a powerful feature!

assets/js/app/DataLayer.js
```diff
@@ -112,12 +112,13 @@ LocusZoom.PositionsDataLayer = function(){
 
     LocusZoom.DataLayer.apply(this, arguments);  
     this.id = "positions";
-    this.fields = ["id","position","pvalue","refAllele","ld:state"];
+    this.xfield = "position";
```
Frencil: Datalayer.js hasn't had any variables that are axis/dimension specific, but [Panel.js](https://github.com/statgen/locuszoom/blob/master/assets/js/app/Panel.js) has and they're all of the form `x_foo` or `y_bar` (or `y1_bar` since they tend to refer to axis-related values at the panel level). Examples include `x_axis`, `x_axis_label`, `x_extent`, and `x_scale`. I'd recommend we stick with underscore spacing in variable names where a dimension and an attribute are both present for consistency.

assets/js/app/DataLayer.js
```diff
@@ -112,12 +112,13 @@ LocusZoom.PositionsDataLayer = function(){
...
+    this.yfield = "pvalue|neglog10";
```
Frencil: Really liking the pipe syntax.

assets/js/app/DataLayer.js
```diff
@@ -112,12 +112,13 @@ LocusZoom.PositionsDataLayer = function(){
...
+    this.fields = ["id", this.xfield, this.yfield, "refAllele", "ld:state"];
 
     this.postget = function(){
         this.data.map(function(d, i){
             this.data[i].ld = +d["ld:state"];
```
MrFlick: Does this just do a numeric conversion? Was there a time that ld wasn't numeric? We could have a "|numeric" transformation but I think this might not be necessary at all. Thoughts?

Frencil: The unary plus operator for number conversion is, presently, the "fastest and preferred way of converting something into a number" (according to [developer.mozilla.org](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Arithmetic_Operators#Unary_plus)). So yes, essentially all this does is ensure the value is a number type, and is shorter syntax and faster execution than `Number()`. It's also more type-safe than `parseInt` or `parseFloat` as certain values like the empty string ("") will return `NaN` with those parse functions but the unary plus will return `0`.

I wouldn't think a `|numeric` pipe would be necessary since a single character unary operator will always be much faster than parsing out and executing a transformation function.

MrFlick: I wasn't surprised about the way you were doing the conversion so much as it was happening at all. So it was really just about being defensive and forcing a conversion. Because the API should already be returning numeric values.

assets/js/app/DataLayer.js
MrFlick: I've updated to x_field and x_field

MrFlick:
As I started to add in the test I realized a bit more refactoring would be needed. I've decided to add the testing as a separate PR. This PR is ready for review and merging.

PR #30, opened 2016-02-23
New layout syntax

Frencil:
# The Problem

Now starting to look at how end users would integrate the LocusZoom plugin into projects it became apparent that the syntax for defining new "layouts" was unintuitive. Layouts were to be built as new Instance classes using methods and, while possible, it was cumbersome and did not lend itself to clarity or easy documentation.

# Layouts as Objects

This branch introduces the concept of a `layout` object, a serializable object that can be built by any means and provided when an instance is created. This approach allows for easily breaking out attributes to cover any aspect of a layout we wish to be configurable, offering far greater flexibility.

## `Instance` and `Panel` classes

With this change there is no longer a need to have subclasses of Instances or Panels. Each base class is still vital to compartmentalize aspects of a rendered LocusZoom object, and now lend themselves even better to broad generalizations (e.g. "An Instance represents a single LocusZoom graph and all of its elements" or "A Panel controls the axes on a graph").

`DefaultInstance`, `PositionsPanel`, and `GenesPanel` no longer exist.

## `DefaultLayout`

Where before `DefaultInstance` created the default layout by extending the base Instance class, now `DefaultLayout` does the same by defining a layout as an object:

```javascript
LocusZoom.DefaultLayout = {
    width: 700,
    height: 700,
    min_width: 300,
    min_height: 400,
    panels: {
        positions: {
            origin: { x: 0, y: 0 },
            width:      700,
            height:     350,
            min_width:  300,
            min_height: 200,
            proportional_width: 1,
            proportional_height: 0.5,
            margin: { top: 20, right: 20, bottom: 35, left: 50 },
            axes: {
                x: {
                    label_function: "chromosome"
                },
                y1: {
                    label: "-log10 p-value"
                }
            },
            data_layers: {
                positions: {
                    class: "PositionsDataLayer",
                    y_axis: {
                        axis: 1,
                        data: "pvalue|neglog10",
                        floor: 0,
                        upper_buffer: 0.05
                    },
                    color: {
                        function: "numeric_cut",
                        parameters: {
                            breaks: [0, 0.2, 0.4, 0.6, 0.8],
                            colors: ["#357ebd","#46b8da","#5cb85c","#eea236","#d43f3a"],
                            null_color: "#B8B8B8"
                        }
                    }
                }
            }
        },
        genes: {
            origin: { x: 0, y: 350 },
            width:      700,
            height:     350,
            min_width:  300,
            min_height: 200,
            proportional_width: 1,
            proportional_height: 0.5,
            margin: { top: 20, right: 20, bottom: 20, left: 50 },
            data_layers: {
                genes: {
                    class: "GenesDataLayer"
                }
            }
        }
    }
};
```

## Recursive layout pattern

The classes `Instance`, `Panel`, and the base `DataLayer` class all keep a `layout` property that *is* their layout. Where before some of the above classes had a `view` property for layout-related data, such as dimensions or dimension bounds, that's all now stored in `layout`. Any methods that change any of those values on the screen should now be using properties in their respective layout objects to track those values under the hood.

## Label functions

The `Panel` class now has a singleton called `LabelFunctions` for storing universal functions for axis labels that change based on state. This also has a robust suite of unit tests built out and ready to go.

## TODO

Lots!

- [x] Add handling for data layer related layout values
  - [x] Add singleton for data coloring methods
  - [x] Generate color scales based on data layer layout
  - [x] Generate y axis extents and scales based on data layer layout
- [x] Expand tests to include *some* data layer stuff
- [ ] ~~Figure out what to do about Instance tests that are now failing without data sources (why did they work before?)~~
- [ ] ~~Expand tests to include more custom-built layouts instead of just the default layout~~
- [ ] ~~Update plugin usage documentation to reflect new syntax and supported options (big!)~~


Frencil:
**UPDATE:** This branch is now development complete.

This is a large refactor and is by no means finished, but in its current state it is stable and functional. As discussed today among LocusZoom devs we're better off freezing further feature development and refactoring on this branch in favor of review and merging to prevent the scope from growing much further.

assets/js/app/DataLayer.js
```diff
@@ -98,30 +88,89 @@ LocusZoom.DataLayer.prototype.reMap = function(){
...
     }.bind(this));
     return promise;
 };
 
+/****************
+  Color Functions
+  Singleton for accessing/storing functions to apply different color schemes to data sets
+*/
+
+LocusZoom.DataLayer.ColorFunctions = (function() {
```
MrFlick: I know we already talked about this, but the general numeric and categorical formatting probably won't be specific to color. They are more general patterns we would work to apply to other attributes.

assets/js/app/DataLayer.js
```diff
@@ -98,30 +88,89 @@ LocusZoom.DataLayer.prototype.reMap = function(){
...
 
-    this.postget = function(){
-        this.data.map(function(d, i){
-            this.data[i].ld = +d["ld:state"];
-        }.bind(this));
-        return this;
-    };
+    LocusZoom.DataLayer.apply(this, arguments);
+    this.layout = layout;
+    this.fields = ["id", "position", "pvalue|neglog10", "refAllele", "ld:state"];
```
MrFlick: We really need to move away from hard coding any values that are pulled from the data sources. These should be configurable at the layout level or somewhere else.

assets/js/app/Instance.js
```diff
@@ -8,8 +8,8 @@
 
   LocusZoom.Instance Class
 
-  An instance is an independent LocusZoom object. Many instances can exist simultaneously
-  on a single page, each having its own data caching, configuration, and state.
+  An Instance is an independent LocusZoom object. Many such LocusZoom objects can exist simultaneously
```
MrFlick: We really should think about renaming this soon. Maybe a "Viewer" or "Plot". Having an instance of an Instance is just a bit confusing.

assets/js/app/DataLayer.js
Frencil: Agreed - because this branch was already pretty extensive in scope I didn't begin diving into abstracting data layers here. However on the [`abstract_data_layers`](https://github.com/statgen/locuszoom/compare/new_layout_syntax...abstract_data_layers?expand=1#diff-4d12c420a365f5ace0fca5d054d05d6fR276) branch (which branches off of `new_layout_syntax`) this is already implemented.

assets/js/app/Instance.js
```diff
@@ -18,46 +18,57 @@ LocusZoom.Instance = function(id, datasource, layout, state) {
...
     this.svg = null;
 
-    // The _panels property stores child panel instances
-    this._panels = {};
+    // The panels property stores child panel instances
+    this.panels = {};
     this.remap_promises = [];
+
+    // The layout is a serializable object used to describe the composition of the instance
+    this.layout = layout || JSON.parse(JSON.stringify(LocusZoom.DefaultLayout));
```
MrFlick: We should consider allowing partial specification of the layout. You can override only the parts of the default layout you want to change, and everything else is preserved (basically merge the lists/objects with parameter values taking precedence over default values). Or maybe we define helper functions to make it less verbose to specify a functional layout object.

assets/js/app/DataLayer.js
Frencil: Agreed - because this branch was already pretty extensive in scope I didn't begin diving into abstracting data layers here. However on the [`abstract_data_layers`](https://github.com/statgen/locuszoom/compare/new_layout_syntax...abstract_data_layers?expand=1#diff-4d12c420a365f5ace0fca5d054d05d6fR276) branch (which branches off of `new_layout_syntax`) this is already implemented.

assets/js/app/Instance.js
```diff
@@ -18,46 +18,57 @@ LocusZoom.Instance = function(id, datasource, layout, state) {
...
+
     return this;
   
 };
 
-// Set the view dimensions for this instance. If an SVG exists, update its dimensions
+LocusZoom.Instance.prototype.initializeLayout = function(){
+
+    // Set instance dimensions or fall back to default values
+    this.layout.width      = this.layout.width      || LocusZoom.DefaultLayout.width;
```
MrFlick: Actually I guess this is related to my note one line 29. We can either merge the user supplied layout with the default layout once at initialization, or we can use the `user||default` pattern before each time we extract a value. Is there a strong reason to choose one over the other?

assets/js/app/Instance.js
Frencil: Good call... as a corollary to this I'd like to make the layout itself more of a "living" object such that you could change values and reload the layout on the fly. Right now you can do this only by destroying and recreating the instance.

I've been kicking around the idea of building a Layout class. It would need methods to spit out or consume a JSON object to keep the portability we've introduced in this branch, but would also potentially help address this concern of defining/consuming partial layouts or changing layout parameters on the fly by keeping all the associated methods in one place. That's a pretty big iteration on the more basic layout approach taken in this branch, though, so should probably be done in a subsequent branch.

assets/js/app/Instance.js
Frencil: I like "Plot" better than "Viewer" for purely subjective reasons but that's still not perfect (it seems sort of interchangeable with Panel, or put another way: it may not be obvious by names alone that a Plot is a thing that contains Panels which in turn contains Data Layers). It's better than Instance, though.

assets/js/app/LocusZoom.js
```diff
@@ -59,7 +61,7 @@ LocusZoom.populateAll = function(selector, datasource, layout, state) {
 
 // Convert an integer position to a string (e.g. 23423456 => "23.42" (Mb))
 LocusZoom.positionIntToString = function(p){
-    var places = Math.max(6 - Math.floor((Math.log(p) / Math.LN10).toFixed(9)), 2);
+    var places = Math.min(Math.max(6 - Math.floor((Math.log(p) / Math.LN10).toFixed(9)), 2), 12);
```
MrFlick: Was this returning more than 12 at some point? I'm unclear about the change.

assets/js/app/Panel.js
```diff
@@ -13,15 +13,15 @@
...
     this.initialized = false;
     
-    this.id     = null;
+    this.id     = id;
     this.parent = null;
     this.svg    = {};
-    
-    this.view = {
+
+    this.layout = layout || {
```
MrFlick: Again, it might be nice to support partial layout specification. So you could say `{height:300}` and have everything else be default. But actually I guess that no one calls this constructor directly? Is it all done via addPanel? So maybe this comment is irrelevant for this code chunk, but then maybe layout should be required?

assets/js/app/Panel.js
```diff
@@ -66,50 +49,107 @@ LocusZoom.Panel = function() {
...
+    this.initializeLayout();
     
     return this;
     
 };
 
+LocusZoom.Panel.prototype.initializeLayout = function(){
+
+    // Set panel dimensions, origin, and margin or fall back to default values
+    this.layout.width      = this.layout.width      || 0;
```
MrFlick: Yeah, here it's not pulling from DefaultLayout as the default if unspecified like in the other case (Instance.js line 47). I guess we just need to decide which pattern we want to use..

assets/js/app/Panel.js
```diff
@@ -190,16 +230,33 @@ LocusZoom.Panel.prototype.initialize = function(){
...
+        throw "Invalid data layer id passed to LocusZoom.Panel.prototype.addDataLayer()";
+    }
+    if (typeof layout !== "object"){
+        throw "Invalid data layer layout passed to LocusZoom.Panel.prototype.addDataLayer()";
+    }
+    if (typeof this.data_layers[layout.id] !== "undefined"){
+        throw "Cannot create data layer with id [" + id + "]; data layer with that id already exists";
     }
-    var data_layer = new DataLayerClass();
+    if (typeof layout.class !== "string" || typeof LocusZoom[layout.class] !== "function"){
```
MrFlick: I'm a bit worried about looking on the main LocusZoom object for resolving layout classes (`LocusZoom[layout.class]`) Especially if a user wanted to add their own they could over right a function we need for something else? Perhaps we could use another custom collection class here with add/get/set methods.

MrFlick:
So it doesn't look like there's actually a "serialize" method yet. Is the idea just to keep the "layout" property/object always current and have no state outside that object?

MrFlick:
Lots of great stuff here. This will be immensely useful.

assets/js/app/LocusZoom.js
Frencil: It looks like this change originated in [commit c97fdc9](https://github.com/statgen/locuszoom/commit/c97fdc995a1aaa5faa79396c5f316120e7a1db8f#diff-8c4275047bc760f446e4203cd9f5e5fdR64). I'm trying to figure out why I made this change, especially considering no associated tests were changed with it. My best guess would be that this function is not doing any type checking or integer casting on the input value... it'll just pump whatever it gets through the math one-liner. One could produce a number greater than 12 by supplying a small enough float value (e.g. 0.0000009 or smaller) and with the new logic increasingly small floats max out at 12 places (while integers behave exactly the same).

That said, I still don't know why I cared about that at the time. Being safer at handling float values (though the function isn't intended to take those) I'd be more inclined to keep this change, but it is more or less unnecessary and a more "correct" change would probably involve some integer casting on the input.

assets/js/app/Instance.js
MrFlick: Maybe we can change it to "Plot" in another pull request and see how we like it.

assets/js/app/Instance.js
MrFlick: I would actually lean the other way in that layout would only be used to "inflate" the object. Other important state would be maintained in classes/variables that can be serialized on demand. The problem with a live layout is that you have to constantly watch for any property changes. And if you create objects based on values in the layout, you'd have to potentially rebuild them everytime. You've been working with the layout far more actively so if you think the living layout would be better long term, i'll trust your judgement.

assets/js/app/Panel.js
Frencil: This is a complex bit of logic to get right. Compounding factors:

* Layouts are treated as nested objects (e.g. the main Instance/Plot layout passes each panel layout into the Panel classes and those in turn pass the data layer layouts into each Data Layer)
* The way JavaScript handles references those layouts on nested objects are references to slices of the original layout stored on the Plot/Instance
* Layout values can be arbitrary types and may be valid as more than one type (good example would be the `color` value for a data layer... could be a string representation of a color or an object that describes the coloring function to use, associated data and parameters, etc.)

I agree that for the most flexibility and best user experience one should be able to define no layout at all at any level or partial layouts and have the gaps "filled in" by the default layout. To make that a reality I'm still leaning in the direction of a Layout class where we can hang generic versions of those functions to be applied at each nested layer. This could also be a place to have generic functions to re-consume a changed layout to see those changes live, import new partial or complete layouts to merge them with current layout definitions, export layouts as serializable objects, etc.

So what about for this branch? Well we're just starting down this layout path... implementing such a class would be a big refactor on top of another big refactor so belongs in a subsequent branch. For this branch we should probably just make the default patterns consistent enough among each other. I'll get this line updated to pull from the default layout.

Frencil:
Re: the serialize method: see [this comment](https://github.com/statgen/locuszoom/pull/30#discussion_r55048797) for some thoughts about that.

In general I see power in keeping the layout object on the Plot/Instance as a serializable object that's kept current as the layout is changed (good example of this in action right now would be the UI to resize an instance which changes top-level layout width/height directly and redraws the plot from the values in the layout). But we need generic methods for merging partial layouts with a fully-defined layout (that fully defined layout could be the default layout, in the case of initialization, or whatever the current layout is, in the case of on-the-fly layout editing that end users could hook up to their own custom inputs).

In general the layout on the plot/instance should always be an accurate description of what that plot/instance *is*. There's an important distinction between that and the `state` object, which describes the current query parameters for the data being displayed. The only other piece needed to essentially "clone" a LocusZoom are the data sources which presently are still built up as a separate object. This pattern is expressed explicitly in code as `LocusZoom.populate` and `LocusZoom.populateAll` take a selector for a target but then apply `datasource`, `layout`, and `state`.

Because these three are different, but similar, and equally important to build a precise LocusZoom visualization we can probably expect each of them to undergo a fair bit of iteration going forward.

assets/js/app/Panel.js
Frencil: Well, hrm. So I looked at this and it's actually ambiguous how to pull these values form the default layout since the default layout has two panels with potentially distinct values. Should we just refer to one of the two default panels (positions/genes) as the "primary" and pull in default values from there? The pattern here with all 0s (or 1s for the proportional dimensions) are mainly just to make sure those values are defined and are integers.

assets/js/app/Instance.js
Frencil: Yeah, that sounds reasonable. The more I think about them as "Plots" the more it makes sense.

assets/js/app/Instance.js
Frencil: Ultimately layout values need to live somewhere. Many of the basic dimension-like layout values lived on an object on the Instance/Plot or the Panel called `view` which had *exactly* the same morphology as the layout object itself does. So in this branch we've dispensed with that and we are essentially using the layout to "inflate" each class (Instance/Plot, Panel, or DataLayer) but then we also cache what was used to do the inflation on the object for quick reference.

We actually haven't gotten too deep into the live editing of layouts yet... the only example of being able to do that now would be the resize UI. While I agree that we have to watch for property changes if the layout object stored on an Instance/Panel/DataLayer we'll have that problem no matter where we stash that information because all objects in JavaScript are editable from anywhere. So we may as well keep it simple and store them exactly how we consume them.

Finally, if we can build a Layout class that would provide an API for safely making live changes to a layout then we document that and hand-editing values in the layout cached on an object is the same as hand-editing values of any attribute of any object: not directly supported and likely to break things. This isn't a problem we should be too worried about... if someone wants to do that they're just gonna do that. If we give them methods for doing it safely most people will use those methods.

Also worth noting: all of this discussion is in reference to a general pattern we would like to have (layouts that can change on-the-fly from a variety of inputs) but realistically the primary use case of LocusZoom as we know it today and in the near future involves being able to define a highly customizable layout up front that *doesn't need to change in the course of a session*. That's what this branch aims to do, so all of this discussion of on-the-fly-editable layouts, while relevant, is not something that needs to be immediately addressed.

MrFlick: Well, there's the idea what we want to be able to look at different phenotype at the same time somewhat soon. We've talked about doing this by adding panels or data layers on the fly so I think we will need to be ready for this soon (but not immediately). So if we want to add panels, do we do so ny editing a layout object, or do we do so by calling some sort of addPanel. I agree that this doesn't have to happen on this PR before merge, it's just something that isn't just hypothetical.

assets/js/app/Panel.js
MrFlick: Looking at the entire file now, I see that the default layout for a panel should have already been passed though and set in the constructor before this function is called. So my initial comments probably aren't actually relevant.

Frencil:
**UPDATE:** Most of our discussion today has been about directions to take layout patterns, API, and workflow beyond the scope of this branch. I've added a few issues to begin to document these needs for addressing on future branches (see #31, #32, and #33).

MrFlick:
Tracking these discussions as other issues makes total sense. Unless you want to make any other changes first, i'd be happy to merge the PR now.

assets/js/app/Panel.js
Frencil: Good call. I've documented this in #34 - probably out of scope for this branch.

Frencil:
Yeah, I'd say go ahead and merge if you're comfortable. I'll continue to document more issues from our discussion and try to keep future refactors smaller in scope. Thanks!

Issue #34, opened 2016-03-04
Safer method for registering custom DataLayers


Frencil:
# The Problem

As of version 0.3.0 (see #30) DataLayer classes are checked for existence via basic string comparison, implying that custom DataLayer classes need to be directly added to the `LocusZoom` singleton. This could create conflicts in the future as implementers of the plugin overwrite existing attributes of the singleton.

# Proposed Solution

As described by @mflick [here](https://github.com/statgen/locuszoom/pull/30#discussion_r55046414) this could be more effectively handled with a custom collection class here with add/get/set methods pattern as seen elsewhere in the plugin.

Frencil:
This was resolved in v0.3.1 (#35)

PR #35, opened 2016-03-04
Abstract Data Layers

Frencil:
# Overview

This branch follows up on the new layout syntax introduced in #30. Here's what it seeks to do in a nutshell:

* Abstract the `PositionsDataLayer` to being a more general-purpose scatter plot data layer
* Port as many configurable options as possible from being hard-coded in data layer classes to being definable in layouts with clear defaults
* Refactor child data layers to use the same singleton pattern as Label Functions and Scale Functions such that custom data layers can be safely added by end users

## DataLayer type: "scatter"

Being a very ubiquitous visualization for many types of data this abstraction was fairly straightforward. Configurable values include point size and shape in addition to color. Implementation of size and shape uses d3's [svg.symbol()](https://github.com/mbostock/d3/wiki/SVG-Shapes#symbol) built-in, affording us six distinct point shapes right out of the box.

## DataLayer type: "genes"

Being a decidedly specific visualization for a specific type of data the abstraction here was more limited than it was for the `ScatterDataLayer` but is nonetheless a big step forward. Configurable parameters now include the following:

* Track height
* Gene label font size
* Track vertical spacing (vertical space between tracks, essentially)
* Label vertical spacing (vertical space between gene name and the tops of a gene's exons)
* Horizontal padding (minimum space horizontally between genes on the same track)

Before this branch all of these were hard-coded values. Now those same values are hard coded as defaults and are configurable via the layout.

Also worth noting: extra logic was added to consider the size of a gene for collisions with other genes (and subsequent track arrangement) based on the size of the gene *and* the size of its label. Previously collisions were detected only by the size of the gene or a minimum of a hard-coded width of 80 as a catch-all for most labels (though several labels are wider than 80 pixels with genes that may be rendered with less width, leading to labels overlapping).

## ScaleFunctions

Where #30 introduced `ColorFunctions` this branch abstracts those out (mainly with name changes) to `ScaleFunctions`. This singleton still implements the get/set/add pattern defined by other singletons. Its purpose is now more explicitly as a place to hang any function that accepts a value, applies a scale of some kind, and returns another value. Currently only implemented for coloring this could be used for scaling the size of data points or really any other attribute we can think of.

## Grouping Singletons

As of this branch we now have three singletons: `LabelFunctions`, `ScaleFunctions`, and `DataLayers`. These are now all direct properties of the core `LocusZoom` object and are defined in a new **Singletons.js** asset with accompanying test file.

# TODO

- [x] Method to merge thin custom layout definitions with default layout objects
- [x] Set defaults in data layers from the default layout objects (instead of hard-coding)
- [x] Fix new initialization bug where panel clip areas are screwy
- [x] Safer method for registering custom DataLayers (see #34)
- [x] Add some data layer specific tests


assets/js/app/Instance.js
```diff
@@ -264,19 +262,23 @@ LocusZoom.Instance.prototype.mapTo = function(chr, start, end){
...
-    }.bind(this), function(error){
-        console.log(error);
-        this.curtain.drop(error);
-    }.bind(this));
+    Q.all(this.remap_promises)
+        .then(this.mapToComplete)
+        .catch(function(error){
+            console.log(error);
+            this.curtain.drop(error);
+        }.bind(this));
```
MrFlick: I've found that adding a `.done()`  at the end of a Q promise can be helpful for debugging. Even with at `.catch()`, there may be an error in that function and the `.done()` helps to give more useful error messages. Of course It should only be used at the terminal end of a promise chain.

assets/js/app/Instance.js
```diff
@@ -264,19 +262,23 @@ LocusZoom.Instance.prototype.mapTo = function(chr, start, end){
...
 
     return this;
     
 };
 
+// Logic to fire when an instance has finished mapping to a new region
+LocusZoom.Instance.prototype.mapToComplete = function(){
```
MrFlick: I'm having a hard time figuring out what this function is supposed to be for from the name. It's for custom event overriding? Is it part of the public API? Should it be `onMapComplete` for an event?

assets/js/app/LocusZoom.js
```diff
@@ -31,15 +31,15 @@ LocusZoom.addInstanceToDivById = function(id, datasource, layout, state){
...
     
 // Automatically detect divs by class and populate them with default LocusZoom instances
 LocusZoom.populate = function(selector, datasource, layout, state) {
     if (typeof selector === "undefined"){
-        selector = ".lz-instance";
+        throw ("LocusZoom.populate selector not defined");
     }
     if (typeof layout === "undefined"){
-        layout = JSON.parse(JSON.stringify(LocusZoom.DefaultLayout));
+        layout = {};
```
MrFlick: This just looks a bit funny not having the treatment of the `layout` be the same as `state`. I would have thought the merging with the default layout would happen here.

MrFlick:
I think this looks really good. I like how those singleton classes are coming together.

assets/js/app/Instance.js
Frencil: Updated to end with `done()`, good catch

assets/js/app/Instance.js
Frencil: There are conceivable situations where some logic should fire only after a remap is complete. The main use case was for the export branch (#27, still in a holding pattern) that needed to update the serialized/encoded SVG string in the download link any time a remap happened. Another example with tooltips, potentially, would be to redraw any tooltips for previously selected data points. However it's not clear that such things would need to happen only at the end of a remap, so I just pulled out that stubbed method entirely on this branch. It, or something like it, can be reapproached when it's actually needed for something.

assets/js/app/LocusZoom.js
Frencil: Latest commit removes those sanity checks from populate, since the Instance constructor is doing them in a more complete way (merging with defaults and whatnot). Arguably that's where that logic belongs, so now populate is a thinner pass-through and will silently pass and undefined parameters to the Instance constructor to gracefully handle.

assets/js/app/Instance.js
MrFlick: OK. I think I was just reading it wrong :) Rather than "mapTo-Complete" I was reading "map -To - Complete" I didn't understand what mapping was happening.

PR #40, opened 2016-03-22
Tool Tips

Frencil:
# Overview

This branch implements a new **Tool Tips** feature that adds a level of customizable interactivity to a layout. A tool tip is an HTML element created and positioned near a data point as prompted by mouse events on that point.

## Tool tip layout syntax

The following tool tips are defined `LocusZoom.DefaultLayout` object in the `positions` and `genes` data layer definitions, respectively:

```javascript
tooltip: {
    divs: [
        { html: "<strong>{{id}}</strong>" },
        { html: "Ref. Allele: <strong>{{refAllele}}</strong>" }
    ]
}
```
```javascript
tooltip: {
    divs: [
        { html: "<strong>{{gene_name}}</strong>" },
        { html: "Gene ID: <strong>{{gene_id}}</strong>" },
        { html: "<a href=\"http://exac.broadinstitute.org/awesome?query={{gene_name}}\" target=\"_new\">EXAC Page</a>" }
    ]
}
```

These demonstrate the general syntax of defining a tool tip. A tool tip layout definition can specify arbitrary HTML or an array of divs, each specifying arbitrary HTML and (optionally) an ID or a class string.

## Tool Tip Creation and Destruction

`LocusZoom.DataLayer` implements the following generic methods:

* `createTooltip(d, id)` - Append an element to the page and fill it with content as specified by the layout
* `destroyTooltip(id)` - Destroy a tool tip by its ID (as in delete it from the DOM entirely)
* `destroyAllTooltips()` - Call `destroyTooltip()` on every defined tool tip for the data layer
* `positionTooltip(id)` - Set attributes to position the HTML element that is the tool tip on the page and create (or update an existing) sibling element that is the connecting arrow
* `positionAllTooltips()` - Call `positionTooltip()` on every defined tool tip for the data layer

When a tool tip is created the data layer tracks it with an object on the data layer called `tooltips`. There is no limit to how many tooltips a single data layer can have in rendered states at any time. Tool tips are created and tracked by an ID which is assumed to be unique (supplied by the data set). Each tool tip object has three elements:

* `data` - data object for the data point to which this tool tip is attached
* `selector` - d3 selector for the HTML element of the tool tip
* `arrow` - d3 selector for the sibling  HTML element of the arrow that visually "connects" the tool tip to the data point

## Tool Tip Positioning

The default `positionTooltip(id)` method on `LocusZoom.DataLayer` positions a tool tip's top left corner at the site of the event (mouse over or click), with an up/left arrow to visually connect the tool tip to the data point.

Discrete data layer classes have the option of implementing their own `positionTooltip(id)` method. In this branch, both the `scatter` and `genes` data layers do just so that tool tip positioning looks better out of the box.

The position method implemented by `scatter` positions tool tips vertically centered to the left or right depending on which half of a data layer a point appears on, attempting to always keep the tool tip within the visual boundaries of the data layer. At the top and bottom of the data layer there is some extra positioning logic to dispense with vertical centering and push tool tips up or down to stay within the data layer's area.

The position method implemented by `genes` positions tool tips above or below genes, horizontally centered. Extra logic has been added to center the arrow on the visible region of the gene and nudge the tool tip left or right if the gene is close to the sides of the data layer's visual area.

## Persistent State

A necessary architectural feature for this branch was expanding the `state` property. Beforehand `state` was only used at the root of an instance to track chromosome and region values. However `state`, in principle, was conceived as an object to store any values supplied by a user during a session with a LocusZoom instance. This would include "selected" data points - data points that have been clicked on.

In order to allow for data layers to use the `state` property in this way it was given a similar treatment to that which `layout` experienced in [v0.3.0](https://github.com/statgen/locuszoom/releases/tag/v0.3.0) and [v0.3.1](https://github.com/statgen/locuszoom/releases/tag/v0.3.1): namely, the `state` object models state values for all layers in an instance (instance, panel, data layer) and does so by having each layer keep its own `state` object which is a reference to a specific part of the global `state` parameter.

To say that  but more clearly: When a `layout` is parsed the instance has a `layout` and a genes panel has a `layout` that is a reference to *just* the genes panel portion of the top level `layout` (and again for the data layers in that panel). If a value is changed in the genes panel layout through `layout` on the genes panel **or** through `layout` on the instance the change is persisted everywhere, because each layer's `layout` is only a reference to its portion of the root `layout` object. With this branch the same is true for `state`.

# To Do

- [x] Update the `parseFields()` logic to use a safer and more widely used double-curly syntax 
- [x] ~~Make `parseFields()` respect/execute data field transform syntax~~
- [x] Test rendering of UTF-8 characters, javascript, and other oddities in tool tip contents
- [x] Write some tests!

### DIFF
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -1,6 +1,8 @@
 $namespace: "lz";
+$default_black: "rgba(24, 24, 24, 1)";
+$default_black_shadow: "rgba(24, 24, 24, 0.4)";
 $tooltip_arrow_width: "7";
-  
+
 svg.#{$namespace}-locuszoom {
 
   background-color: #FFFFFF;
@@ -8,6 +10,10 @@ svg.#{$namespace}-locuszoom {
   cursor: move;
   font-family: "Helvetica Neue", Helvetica, Aria, sans-serif;
   font-size: 10px;
+
+  rect.#{$namespace}-clickarea {
+    fill: transparent;
+  }
 
   .#{$namespace}-curtain {
     rect {
@@ -40,7 +46,7 @@ svg.#{$namespace}-locuszoom {
 
   .#{$namespace}-html_module-info_box {
     position: absolute;
-    border: 1px solid #181818;
+    border: 1px solid #{$default_black};
     border-radius: 3px;
   }
 
@@ -80,17 +86,17 @@ svg.#{$namespace}-locuszoom {
   }
 
   path.#{$namespace}-data_layer-scatter {
-    stroke: #181818;
+    stroke: #{$default_black};
     stroke-width: 1;
   }
 
   path.#{$namespace}-data_layer-scatter-hovered {
-    stroke: #181818;
+    stroke: #{$default_black};
     stroke-width: 3;
   }
 
   path.#{$namespace}-data_layer-scatter-selected {
-    stroke: #181818;
+    stroke: #{$default_black};
     stroke-width: 4;
   }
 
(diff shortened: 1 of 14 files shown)


MrFlick:
I think the tooltips look great and this method seems quite extensible. Right now the divs are all objects with html attributes. If we wanted to define other block types, would we just set different attributes? Like I could see a table/iterator type pattern being useful. Or some sort of conditional display criteria. It seems like we'd need something more that just literal HTML in order to be able to offer some interaction for changing the LD reference SNP or the conditional SNP for a conditional analysis. Did you already have something in mind for that?

Issue #41, opened 2016-03-22
Make instance resizing automatic, manual, or disabled


Frencil:
Presently by default all instances are manually resizable (with the `resize_handle` UI element).

At a minimum this should be a parameter that can be disabled, forcing an instance to keep a strictly defined size. Ideally a layout could also set this value to be something like "automatic" such that the instance always takes up 100% of its containing div.

Frencil:
Thinking about this, the "automatic" behavior can definitely get tricky. For example, LocusZoom's existing resize logic allows for any arbitrary aspect ratio, so you could have a LocusZoom plot that's many thousands of pixels tall. As a page / containing dev resizes, if a LocusZoom plot is set to automatically stay at 100%, how should it handle changing its vertical size? Just maintain the present aspect ratio, perhaps?

@MrFlick / @benralexander thoughts?

benralexander:
I agree that maintaining the aspect ratio would be a good solution.

On Tue, Mar 22, 2016 at 3:52 PM, Christopher Clark <notifications@github.com

MrFlick:
It seems like page resizing should just effect the width of the plot. Changing the height makes sense as a drag. Right now we have drag in the order to change both width and height. Maybe just allow it to change the height. Since we're not really making plots where the distance on the x and y axis are directly relatable, i'm not sure it makes sense to try to preserve aspect ratios. That's more important and x and y values are on the same scale.

benralexander:
To be fully integrated with the rest of the portal there probably shouldn't
be an independent drag event, and instead the graphic should drive its
width from the page size.  At small screen sizes I could imagine that a
presentation with a preserved aspect ratio would be more readable than one
that is tall but skinny.  That said, I could also imagine someone
intentionally making the screen very wide in order to better examine
genomic position, and not necessarily wanting the graphic to become
taller.  So in the end I'm not sure in the end whether preserving the
aspect ratio will be better or worse, but I imagine it's not a big coding
switch either way (i.e. let's see how it feels once we get it in the
portal).  I do think, however, that we want to make the LZ's width (at
least) dependent on browser page size, and not independently manipulable
via a corner drag.

Frencil:
Thanks for the discussion on this, folks. Here's an update...

A working prototype for responsive resizing is now up in #42 (working demo [here](http://portaldev.sph.umich.edu/lzplug_staging/demo_responsive.html)). Things to note:

[...]

PR #42, opened 2016-03-23
Responsive Resizing

Frencil:
# Overview

This branch is built to address #41. In a nutshell, a layout should support making a plot discretely sized with no ability to resize, manually resizable (by way of a UI element), or *responsively* resizable (in that it expands to fill its available space).

### `resizable`

The `resizable` parameter is defined at the instance/plot level. This can be "manual", "responsive", or null/false.

### `aspect_ratio`

For resizable plots, since container width is typically bounded by window width but height is theoretically infinite, an `aspect_ratio` value can be defined to keep a specific aspect ratio for the plot as it resizes. This is a numerical value defined as width / height. So, for example, a square aspect ratio would be `1` and a 16:9 aspect ratio would be `1.7777777` (16/9).

### `proportional_origin`

Another other major change is that panels now have `proportional_origin` in addition to `proportional_width` and `proportional_height`. In general the proportional geometry values are used whenever a layout is resizable by either method, because at that point predefined discrete dimensions and origins may only be valid when the plot is first loaded (and in the case of a responsive plot, may immediately be overwritten during initialization). The `proportional_origin` is similar to the `origin` in that has `x` and `y` parameters, but these are expressed as multipliers applied to the dimensions of the plot. So, for example, this:

```javascript
proportional_origin: { x: 0.25, y: 0.5 }
```

Would position a panel 1/4 of the plot width from the left edge of the plot and 1/2 of the plot height from the top of the plot. As the plot is resized (by either method) the proportional origin would stay constant while the discrete origin would likely change.

## CSS

Responsive plots really only make sense inside a containing element that has the right CSS to expand to fill space. A new namespaced class has been added to make this easier for implementers: `lz-container-responsive`.

## Demo Responsive

Finally, I reworked the existing demo.html into a new demo_responsive.html (this exists in addition to the original). This demo page demonstrates how to build a responsive LocusZoom plot and applies some conventions that ultimately would be good to apply to the original demo as well (for example: clear separation of LocusZoom config code from demo-specific logic, illustration of applying a partial layout, etc.)

### DIFF
--- a/assets/js/app/Instance.js
+++ b/assets/js/app/Instance.js
@@ -77,10 +77,10 @@ LocusZoom.Instance.prototype.setDimensions = function(width, height){
     // Set the layout height to a discrete value, by passed argument or responsive logic using the aspect ratio
     if (!isNaN(height) && height >= 0){
         this.layout.height = Math.max(Math.round(+height), this.layout.min_height);
-    } else if (this.svg && this.layout.resizable == "responsive"){
+    } else if (this.layout.resizable == "responsive"){
         this.layout.height = this.layout.width / this.layout.aspect_ratio;
     }
-    // Keep aspect ratio in agreement with width and height
+    // Keep aspect ratio and dimensions in agreement
     this.layout.aspect_ratio = this.layout.width / this.layout.height;
     // Apply layout width and height as discrete values or viewbox values
     if (this.svg != null){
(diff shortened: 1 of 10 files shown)


assets/js/app/Instance.js
```diff
@@ -43,6 +46,14 @@ LocusZoom.Instance = function(id, datasource, layout, state) {
 
 LocusZoom.Instance.prototype.initializeLayout = function(){
 
+    // If this is a responsive layout then set a namespaced/unique onresize event listener on the window
+    if (this.layout.resizable == "responsive"){
+        this.window_onresize = d3.select(window).on("resize.lz-"+this.id, function(){
```
Frencil: Note that this is a "namespaced" event listener. I didn't even know you could do this until today, but essentially the ".lz-" and the id following the "resize" still ties this method to a resize event on the window but does *not* affect any other existing resize event listeners.

Frencil:
UPDATE: this just got some tests added and is essentially complete. A quick review pass would be most helpful but isn't necessary. I'll merge it in later today to make way for Tool Tips (#40) to be on deck for final testing/review/merge early Friday.

Issue #46, opened 2016-04-02
Layout Builder Demo Page


Frencil:
Presently we have two demo pages in the repo to demonstrate setting up a LocusZoom plugin with only a slight variation in the layout. Moving forward we're going to be supporting an increasing amount of flexibility in the layout with only one configuration baked into `LocusZoom.DefaultLayout`, relegating all that flexibility to only the [API Reference](https://github.com/statgen/locuszoom/wiki/API-Reference).

By building a dynamic demo page that has form elements to tweak the layout (which evolves with the library) we'll not only have a tool to demonstrate the level of variation LocusZoom affords developers but also have something we can stand up on [locuszoom.sph.umich.edu](http://locuszoom.sph.umich.edu/locuszoom/) to highlight current LocusZoom library development.

A minimum viable product might include:

* A couple of big text areas showing the current layout and state objects
* Form controls to submit any changes to the layout/state and apply them to the plot
* A link to the API reference

It may be wise to avoid creating a big dynamic HTML form without first figuring out a way to model all layout possibilities in some sort of portable data structure which the library can also use for layout verification. Otherwise all new layout rules will need to be implemented twice and there will be a high risk of divergence between the library and the layout builder demo.

PR #47, opened 2016-04-05
Unified Layout with Plot Builder

Frencil:
# What

This branch was built to address the idea put forth in #46 - a tool to be able to see an editable layout and plot side by side as an interactive Layout/Plot builder.

In building this tool a few latent bugs in initialization, layout parsing, and state I/O quickly surfaced. This led to an idea that's been brought up before coming to pass - **merging the layout and state**.

## Unified Layout

Wherever there once was a layout and/or a state, now there's just a layout. State is still there. It's just *inside* layout, and it's still called `state`.

From this the following changes fall out:

* Initialization functions like `LocusZoom.populate()` **still take `state` as a fourth argument for backward compatibility**. Now, however, `populate()`'s logic merges any state passed in with state defined in the passed in layout and state detected from the `data-region` parameter on the container into a single `state` object in the layout.  

* The nested structure of panels and data layers is no longer repeated in both layout and state.  

* A new state syntax has been developed for the states of panels and datalayers to allow for tracking their state in a deterministic way without building a deeply nested JSON object (the thinking here is to make it easy for implementers to pre-load state).

## Plot Builder

A new demo appears in this branch - `plot_builder.html`. This is a simple tool that shows a plot next to a textarea filled with the current layout with simple UI and hooks to keep them in sync with one another as either is changed.

### onUpdate()

To make the tool function the plot builder needed a means to know when the layout has been updated (e.g. the plot resizing, any data query, selecting elements in the plot, etc.) so this branch adds support for a user-defined `onUpdate()` event handler that can be added to a plot. It is automatically triggered whenever a plot's layout is effectively changed.

## Status

This branch is now **Development Complete**.

The attention paid to backward compatibility should mean that existing LocusZoom implementations are not disrupted, unless those implementations are directly passing state objects to pre-select points on data layers. That syntax has been simplified in this branch, so that change will need to be communicated.

### DIFF
--- a/assets/js/app/LocusZoom.js
+++ b/assets/js/app/LocusZoom.js
@@ -3,12 +3,12 @@
 /* eslint-disable no-console */
 
 var LocusZoom = {
-    version: "0.3.6"
+    version: "0.3.7"
 };
     
 // Populate a single element with a LocusZoom instance.
 // selector can be a string for a DOM Query or a d3 selector.
-LocusZoom.populate = function(selector, datasource, layout, state) {
+LocusZoom.populate = function(selector, datasource, layout) {
     if (typeof selector == "undefined"){
         throw ("LocusZoom.populate selector not defined");
     }
@@ -23,7 +23,7 @@ LocusZoom.populate = function(selector, datasource, layout, state) {
             this.attr("id", "#lz-" + iterator);
         }
         // Create the instance
-        instance = new LocusZoom.Instance(this.node().id, datasource, layout, state);
+        instance = new LocusZoom.Instance(this.node().id, datasource, layout);
         // Add an SVG to the div and set its dimensions
         instance.svg = d3.select("div#" + instance.id)
             .append("svg")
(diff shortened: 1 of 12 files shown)


assets/js/app/LocusZoom.js
```diff
@@ -3,25 +3,40 @@
...
 // selector can be a string for a DOM Query or a d3 selector.
-LocusZoom.populate = function(selector, datasource, layout, state) {
+LocusZoom.populate = function(selector, datasource, layout) {
     if (typeof selector == "undefined"){
         throw ("LocusZoom.populate selector not defined");
     }
+    // Empty the selector of any existing content
+    d3.select(selector).html("");
+    // If state was passed as a fourth argument then merge it with layout (for backward compatibility)
+    if (typeof state != "undefined"){
```
MrFlick: How would `state` ever get set here if it's been removed as a named parameter to the function? Shouldn't this be an undefined variable error?

Frencil: My mistake - I had removed it from the arguments list on `populate` and `populateAll` originally on this branch but set out to restore it per our recent conversation. I added the processing of a state argument but just forgot to add it back to the arguments list.

assets/js/app/LocusZoom.js
```diff
@@ -48,7 +59,7 @@ LocusZoom.populate = function(selector, datasource, layout, state) {
 LocusZoom.populateAll = function(selector, datasource, layout, state) {
     var instances = [];
     d3.selectAll(selector).each(function(d,i) {
-        instances[i] = LocusZoom.populate(this, datasource, layout, state);
+        instances[i] = LocusZoom.populate(this, datasource, layout);
```
MrFlick: State should be passed through here as well. Perhaps add a test for this behavior? Maybe throw a warnings that this parameter is being deprecated?

PR #48, opened 2016-04-08
Modify GeneSource to allow for positional transcript annotation

MrFlick:
This is an extension of the gene data source to allow for annotation of positions indicating weather or not a position is in the exon of a transcript or not.

Now you can request a field from the gene data source like `gene:transanno(ENST00000277945.7)` and that will return 1 for each position that is in that transcript but not in an exon, 2 for each position in an exon of that transcript, and 0 otherwise.

The basic logic for the merging can be tested with

```
var gs = new LocusZoom.Data.GeneSource("http://foo.bar")
var mockResp = {data:[{gene:1, transcripts:[]},{gene:2, transcripts:[{transcript_id:"ENST1", start: 2, end:15, exons:[{start:2, end:3}, {start:7, end:12}]}]}]};
var mockChain = {body:[{position:1},{position:5}, {position:10}, {position:30}]};
gs.parseResponse(mockResp, mockChain, ["transanno(ENST1)"],["anno"])
```

With the demo page data_sources object ...

```
var req = new LocusZoom.Data.Requester(data_sources);
req.getData({chr:10, start:114550452, end:115067678},["position","pvalue","gene:transanno(ENST00000277945.7)"]).then(function(x) {console.log(x)}).done()
```

you should see that the 211st element has a value indicating inclusion in the gene

### DIFF
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -274,8 +274,71 @@ LocusZoom.Data.GeneSource.prototype.getURL = function(state, chain, fields) {
         " and start le " + state.end +
         " and end ge " + state.start;
 };
+LocusZoom.Data.GeneSource.prototype.findTranscriptPositions = function(data, transName) {
+    var transData;
+    for(var g=0; g < data.length && !transData; g++) {
+        for (var t=0; t < data[g].transcripts.length; t++) {
+            if (data[g].transcripts[t].transcript_id==transName) {
+                transData = {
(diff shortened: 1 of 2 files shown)


abought:
Checking in on old PRs; I notice this one has developed merge conflicts.

No rush, but is this feature still something we want / need? Is it worth bringing up to date?

abought:
Closing stale PR, but very open to revisiting if there is still a need.

PR #49, opened 2016-04-20
Expand Tick Controls

Frencil:
# What

In looking at expanding the LocusZoom platform to handle PheWAS plots it became clear that finer control over the positioning and appearance of axis ticks would be necessary. This branch iterates over how axis layouts can be built and interpreted to provide this new level of fine-grain control.

## Details

Ticks actually used by a panel are no longer stored in the layout but on the panel object itself. This opened up the ability to define discrete ticks in the layout and have them interpreted into the panel object.

Discrete tick definitions in the layout can be done by specifying x offsets, style objects, and transform objects to create arbitrarily many ticks with individual control over color and geometry, allowing for LocusZoom to support the PheWAS-style x tick pattern.

Two new values have also appeared in the `StandardLayout` to preserve consistent rendering:

* `panel.axis.{x|y1|y2}.label_offset`  
  A value in pixels to position an axis label relative to its axis. Previously all axis labels were positioned by adding/multiplying magic numbers to layout margins, so chosen to look good for the narrow use case of just the StandardLayout. This change remodels those values as discrete pixel values with defaults defined in the StandardLayout.  

* `panel.axis.{x|y1|y2}.tick_format`  
  A special value that defaults to "region" for the x axis in the positions panel in `StandardLayout`. This breaks the previous pattern where all x axis ticks were being piped through `LocusZoom.positionIntToString()` as if they were always position values, and instead makes that behavior an optional bit of logic that can be triggered from the layout.

### DIFF
--- a/assets/js/app/DataLayer.js
+++ b/assets/js/app/DataLayer.js
@@ -62,7 +62,7 @@ LocusZoom.DataLayer = function(id, layout, parent) {
             selector: d3.select(this.parent.parent.svg.node().parentNode).append("div")
                 .attr("class", "lz-data_layer-tooltip")
                 .attr("id", this.getBaseId() + ".tooltip." + id)
-        }
+        };
         if (this.layout.tooltip.html){
             this.tooltips[id].selector.html(LocusZoom.parseFields(d, this.layout.tooltip.html));
         } else if (this.layout.tooltip.divs){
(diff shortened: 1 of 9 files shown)


MrFlick:
Looks good to me. Happy to merge. Did you want to take a pass @pjvandehaar ?

Frencil:
Thanks for the review, @MrFlick. I'm going to go ahead and merge since I've got another branch off of this one ready for review.

PR #50, opened 2016-04-26
Static Data and the Line Data Layer

Frencil:
This branch implements the minimum viable product to render a line of genome-wide significance on positions panel. It does so by implementing a basic data layer of type `line`, as well makes use of the `StaticJSON` data source introduced in #53.

## Line Data Layer

The `line` data layer uses d3's `line()` method to render a standard line. Layout parameters supported include `style`, and `transition`, as well as the line-specific parameter of `interpolate` which is passed directly through to d3's `interpolate` function.

Selectability is not currently implemented for the line data layer, as it's not clear what that behavior should be.

Tool tips have been implemented on this branch with a custom positioning function. What's more, to improve the "tightness" of the UI's feel, line data layers with defined tool tip content will actually render the path twice. The first pass will take styles from the layout and be visible; the second pass will render an identical invisible line on top with a predefined stroke width of `12px`. This second line acts as the "hit area" for mouse events. All mouse events related to tool tips use this line instead of the rendered line but both lines are joined to the same data underneath, resulting in the ability to have thin or dotted/dashed lines that have wide, continuous mouse hit areas.

## To Do

Open question which of these are in scope for this branch or could be added later.

- [x] Tool tips
  - [x] Basic mouse event binding
  - [x] Line-data-layer-specific tool tip positioning methods
- [x] Tests
  - [x] Data layer axis minimum extent logic

### DIFF
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -103,6 +103,12 @@ svg.#{$namespace}-locuszoom {
     stroke-width: 4;
   }
 
+  path.#{$namespace}-data_layer-line {
+    stroke: #{$default_black};
+    stroke-width: 1;
+    cursor: pointer;
+  }
+
   g.#{$namespace}-data_layer-gene {
     cursor: pointer;
   }
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -109,6 +109,12 @@ svg.#{$namespace}-locuszoom {
     cursor: pointer;
   }
 
+  path.#{$namespace}-data_layer-line-hitarea {
+    stroke: transparent;
+    stroke-width: 12px;
+    cursor: pointer;
+  }
+
   g.#{$namespace}-data_layer-gene {
     cursor: pointer;
   }
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -110,6 +110,7 @@ svg.#{$namespace}-locuszoom {
   }
 
   path.#{$namespace}-data_layer-line-hitarea {
+    fill: transparent;
     stroke: transparent;
     stroke-width: 12px;
     cursor: pointer;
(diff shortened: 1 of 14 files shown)


MrFlick:
This `static_data` really doesn't feel like it belongs in the base DataLayer. Do we not have different data layer classes? I had assumed that there were sub-classes for different types like scatter plots or line plots but it seems we are just decorating instances of the base DataLayer for each of the layer types. Each layer should have the option of overriding how it gets or sets its own data if it likes, with the default behavior being to request fields from the data requester. The behavior seems like it should be specific to the specific data layer. That way those properties can have friendlier names.

It would be nice to follow a pattern with significance lines similar to the `abline` function in R where you can set a horizontal line value, or a vertical line value (or, less useful in this case, an intersect and slope). Having to hard code "magic" extreme values seems unfriendly.

Frencil:
> This static_data really doesn't feel like it belongs in the base DataLayer. Do we not have different data layer classes? I had assumed that there were sub-classes for different types like scatter plots or line plots but it seems we are just decorating instances of the base DataLayer for each of the layer types. Each layer should have the option of overriding how it gets or sets its own data if it likes, with the default behavior being to request fields from the data requester. The behavior seems like it should be specific to the specific data layer. That way those properties can have friendlier names.

Given that JavaScript isn't really a true class-based language our Data Layer subclass pattern is about as subclass-y as JavaScript really gets... each "subclass" applies the Data Layer prototype and redefines only what's necessary. This tends to involve a subclass-specific default layout, a subclass-specific render function, and usually a subclass-specific function for placing tool tips (as each type of data may require tool tips be positioned differently, though composed by the same rules).

As a first pass this is a simple implementation of static data, as a concept. If defined the data layer uses it, if not defined it goes through the standard getData approach. To split this out to each data layer on this branch would not only require copying that logic three times, but having each data layer subclass implement its own `reMap()` method. That seems like a lot of overhead and code duplication. I would also argue that overriding data requests with static data is not something that is unique to a given data layer type. In the general sense there's no reason to make that something that, say, only the line data layer is capable of, when the layout syntax would be the same regardless.

Another thing to consider is where is that behavior defined? Would we require an implementer to reimplement `reMap()` functions on data layer subclasses in order to use the static data pattern or something like it? It makes sense to ask the implementer to write a bunch of functional javascript to make their data sources work because those could be just about anything, but the idea behind static data is that when you have a use for it you add a data layer specifically for it. It would be for more architecturally complex to somehow combine static and dynamic data in the same data layer when it's much easier to just add another data layer and keep a separation between those that are static and those that are dynamic. Future iterations of LZ will require more control in the plugin over the ordering of data layers (e.g. changing the z-index arrangement, toggling visiblity, etc.) and by establishing a pattern where static/dynamic data just don't mix on the same data layer we get a simpler workflow to follow for end users.

> It would be nice to follow a pattern with significance lines similar to the abline function in R where you can set a horizontal line value, or a vertical line value (or, less useful in this case, an intersect and slope). Having to hard code "magic" extreme values seems unfriendly.

I could see the argument here. This sounds, however, like a feature with a more specific set of use cases (e.g. draw straight lines, as opposed to drawing any fixed data for any data layer). They aren't mutually exclusive but what's in this branch is more generally applicable and covers the straight line use case with some syntax that is more direct. I agree that directness can make for less friendly syntax - I need to know something about the domain of my data to draw a straight line that spans it, whereas saying "draw a horizontal line at some y value" would effectively generate the x values from the domain under the hood.

In that way I see this more general approach as a precursor step to that more specific approach and others like it. It allows for all the same functionality but with different syntax that requires a bit more context knowledge up front. I could see adding syntax specifically for straight lines using R's `abline` as a pattern in a future iteration, but would only consider that a worthwhile feature in addition to the more general static data pattern, not as a replacement for it.

Frencil:
**UPDATE** - with #53 having merged I've ported this branch to no longer define static data in the layout and instead use the `StaticJSON` data source.

The line of significance is still being created by adding a data layer and tying it to a StaticJSON data source with hard-coded data to generate a line from two points. While this works it's not ideal... as previously discussed elsewhere it would probably make more sense to allow for the creation of simple graph elements like lines of significance in the layout as a property of the panel (e.g. with syntax more in the pattern of "horizontal line at y" or perhaps borrowing from R's [abline](https://stat.ethz.ch/R-manual/R-devel/library/graphics/html/abline.html) function).

However we move forward on that I'd like to get a better sense of what types of lines may need to be generated independently of data like that to form a more general approach. It would also extend the scope of this branch pretty far, so I would be inclined to call this branch feature complete and tackle a better static line feature in a future iteration.

PR #53, opened 2016-05-09
Data Source Updates

MrFlick:
This request adds some new features to the data sources for the API.

First, it incorporates the new Recombination Rate API endpoint. On the demo page, this data can be accessed via

     demo_instance.lzd.
          getData(demo_instance.state, ["recomb:position", "recomb:recomb_rate"]).
          then(console.log.bind(console)).done();

Secondly, there is now a data class for Static data. You can add static sources like this

    data_sources.addSource("hello", ["StaticJSON", {a:[1,2,3], b:[4,5,6], c:[7,8,9]}])
    data_sources.addSource("bye", ["StaticJSON", [{a:1, c:7},{a:2, c:8},{a:3, c:9}]])

and then retrieve the data with

    demo_instance.lzd.
        getData(demo_instance.state, ["hello:a", "hello:c|neglog10"]).
         then(console.log.bind(console)).done();
    demo_instance.lzd.
         getData(demo_instance.state, ["bye:a", "bye:c|neglog10"]).
         then(console.log.bind(console)).done();

This method accepts either arrays of named objects, or a named object of arrays. It allows for the application of transformations as indicated by this example

Finally, the code required to create a new Data Source has been streamlined. There is now the

    LocusZoom.Data.Source.extend()

function which takes a constructor (and an optional unique ID for serialization) and sets up all the prototypes correctly for the object. There is now also a default `parseResponse()` method that takes care of copying over names with the proper namespace and applying transformations. This should make it much easier to create custom data sources.


assets/js/app/Data.js
```diff
@@ -9,7 +9,13 @@ LocusZoom.DataSources = function() {
     this.sources = {};
 };
 
+LocusZoom.KnownDataSources = [];
```
Frencil: Having `KnownDataSources` and `DataSources` as separate (but tightly coupled) attributes of the core `LocusZoom` object isn't as clean as it could be. A pattern used in other places (`Transformations`, `LabelFunctions`, `DataLayers`, etc.) is to have a single object in the core that contains a structure for storing members as well as whatever methods are necessary for adding/deleting/interacting with members:

```javascript
LocusZoom.DataSources = (function() {
  var obj = {};
  var sources = {};

  obj.add = function(name, data_source) {
    ...
  };

  obj.set = function(name, data_source) {
    ...
  };

  ...

  return obj;
})();
```

A minor bonus to this is that the `DataSources` object should know what it does and does not have, so there's no need for checks against the existence of `KnownDataSources` as appears in `findKnownSource()`.

MrFlick: I agree that data sources should probably become a standard singleton like the rest. I can move that (though it will introduce breaking changes if we change the add/set method names. I can add an extra method with a depreication warning.).

But that still wound't eliminate the KnownDataSources collection. This is the collection of all defined data sources (not necessarily the ones you are using). So we include in the library an object to represent our Recombination Rate API endpoint. That exists whether or not you use it or not. (If you do want to use it, you add it to your DataSources collection). Are you handling this distinction differently with the data layers?

With this current PR, one should never have to use the KnownDataSources collection directly, but can if they need to (if they don't want to use the new extend() method to create custom data sources but still want all the other advantages of deserialization).

assets/js/app/Data.js
```diff
@@ -175,30 +178,94 @@ LocusZoom.Data.AssociationSource.prototype.parseResponse = function(resp, chain,
...
+LocusZoom.Data.Source.prototype.parseObjectsToObjects = function(x, fields, outnames, trans) {
+    //intended for an array of objects
+    // [ {"id":1, "val":5}, {"id":2, "val":10}]
+    var records = [];
+    var fieldFound = [];
+    for (var i = 0; i < x.length; i++) {
+        var record = {};
+        for (var j=0; j<fields.length; j++) {
+            var val = x[i][fields[j]];
+            if (typeof val != "undefined") {
```
Frencil: It looks like `fieldFound` is only getting an index populated with 1 if the field at that index `j` is found. Thus if I've got a list of four fields: `["a", "b", "c", "d"]` and "c" isn't found, `fieldFound` would be `[0, 1, 3]`. At the end of the function in the `forEach` on line 203, then, `v` would always be "1" and an exception would never be thrown for a missing field. Should this conditional have an `else` that sets `fieldFound[j]` to `false` or `0` to get the desired effect?

assets/js/app/Data.js
Frencil: Currently with all other singletons there's no distinction between the singleton members LocusZoom defines in the core, the members you as a user define, and the members you're actually using. From the perspective of Data Layers, Transformations, Scale Functions, etc. I'm not sure such a distinction really matters... there's just one collection that you can add to (or delete from at your own risk) and use whatever you need to whenever you need to. This simplifies the storage of such members as there's only one place they live, and affords future flexibility to begin (or stop) using members of any singleton well after initialization (e.g. as triggered by a UI element or something).

locuszoom.app.js
```diff
@@ -437,13 +452,17 @@ LocusZoom.DataSources.prototype.getSource = function(ns) {
     return this.sources[ns];
 };
 
+LocusZoom.DataSources.prototype.removeSource = function(ns) {
+    return this.setSource(ns, null);
+};
+
 LocusZoom.DataSources.prototype.setSources = function(x) {
```
Frencil: Possibly out of scope for this branch, but it looks like `setSources` is only used in the testing layer? Is there a use case outside of testing where this method would be useful? If not I'd recommend it be defined in the test suite directly to keep the core restricted to logic used in the direct execution of the plugin only.

locuszoom.app.js
```diff
@@ -674,19 +737,38 @@ LocusZoom.Data.GeneSource.prototype.getURL = function(state, chain, fields) {
...
+
+LocusZoom.Data.RecombinationRateSource.prototype.getURL = function(state, chain, fields) {
+    var source = state.recombsource || chain.header.recombsource || this.params.source || 15;
+    return this.url + "?filter=id in " + source +
+        " and chromosome eq '" + state.chr + "'" + 
+        " and position le " + state.end +
+        " and position ge " + state.start;
+};
+
+LocusZoom.Data.StaticSource = LocusZoom.Data.Source.extend(function(data) {
```
Frencil: Thanks for setting this up. We can make this branch a dependency for #50 so that branch can use this pattern for defining static data instead of using the layout.

assets/js/app/Data.js
MrFlick: Good point. Actually. If `c` isn't found, it should be `[1,1,undefined,1]` but `forEach` seems to skip over undefined values. It would have been even worse if the missing field(s) was(were) at the end of the list. I'll go ahead and initialize all to 0.

locuszoom.app.js
MrFlick: The use case is basically serialization (which we aren't really using yet in our simple demo page). It allows you to set all the sources at once with a JSON object (or string). It's the inverse of the `JSON.stringify()` for the `DataSources` collection.

assets/js/app/Data.js
MrFlick: Each locuszoom instance can/should have it's own DataSources collection. The KnownDataSources are meant to be universal. It's the list of recipes while DataSources is a realization of a particular meal (where each dish is given some unique name). I don't see a logical way which they can be combined. Maybe I just need better names or some other level of abstraction.

Frencil: I think this pattern of a single object with methods, storage, and initial recipes all included still works, it just wouldn't behave as a singleton per se. The workflow for the implementer would still be to instantiate a new object a la `var mydatasources = new LocusZoom.DataSources();`, and then they can add/modify/remove as they see fit by using methods on the `mydatasources` object. When ready `mydatasources` gets passed into the populate method to be specific to the instance that's then created. By this pattern each LocusZoom instance would retain its own copy of the built-in and whatever customized data sources it was given at the populate step, and that may include slight modifications to built-in sources from instance to instance.

Frencil:
All looks good! Merging.

PR #55, opened 2016-05-19
Recombination Rate Line

Frencil:
# What

This branch utilizes the new `recomb` data source to implement a basic recombination rate line.

While the line data layer already existed the tool tip positioning logic needed some iteration to behave a bit more intuitively for all lines in general.

## Changes

* Line tool tip positioning now takes place to the left/right of the line if the slope is greater than one (including `Infinity`) and to the top/bottom of the line if the slope is less than one.
* Abstract tool tips new have a flow of `createTooltip()` => `updateTooltip()` => `positionTooltip()` (as opposed to just `createTooltip()` => `positionTooltip()`). The new `updateTooltip()` function handles writing the contents to the tool tip. This doesn't change the behavior for tool tips on other data layers that create tool tips once for discrete data points, but lines are continuous data ranges so the update method allows for changing the contents of the tooltip as it is repositioned (e.g. as the mouse moves over the line - useful if the tooltip is showing dynamic data about that position on the line).
* The `StandardLayout` now includes a recombination rate data layer in the positions panel, arranged behind the scatter/association data but in front of the line of statistical significance.
* Tool tips for recombination rate are *not* enabled in the standard layout. This is because while tool tips for lines are functional they're still a bit clunky.

### DIFF
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -119,9 +119,8 @@ svg.#{$namespace}-locuszoom {
   }
 
   path.#{$namespace}-data_layer-line-hitarea {
-    fill: transparent;
+    fill: none;
     stroke: transparent;
-    stroke-width: 12px;
     cursor: pointer;
   }
 
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -269,6 +269,7 @@ div.#{$namespace}-data_layer-tooltip-arrow_bottom_right {
   font-family: "Helvetica Neue", Helvetica, Aria, sans-serif;
   font-size: 80%;
   padding-top: 4px;
+  padding-bottom: 4px;
   .#{$namespace}-controls-button {
     text-decoration: none;
     padding: 0.2em 0.5em 0.2em 0.5em;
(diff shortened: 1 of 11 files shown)


MrFlick:
I noticed this only works for `demo_responsive.html`, not `demo.html`. Should we still keep the two separate should we merge down to one demo?

For the recombination rate label, we should probably include the units "cM/Mb" aka centimorgans per megabase. Otherwise it's less clear what those values actually mean. If that looks too bad, we may have to thing of other ways to label that access. But it's so nice to see that friendly blue line there again.

One other minor thing, the Download SVG button seems to be getting cut off. Screenshot attached.

image

Other than that, things look great!

Frencil:
Issues addressed:

* `demo.html` now correctly defines the recomb data source to display correctly
* The axis label now has units (cM/Mb) and the panel in the standard layout has a minimum height to ensure room for the full axis label
* The controls bar now has a bottom padding to prevent cutoff of elements inside

As for merging the demos, I was thinking that as LocusZoom's capabilities continue to expand we may want the repo to host a variety of demos (e.g. we could host a demo of using LocusZoom for a PheWAS plot right now, but that's not exactly high priority). I could see migrating all demos into a `demos` directory or some such and renaming / updating them to highlight specific features of the library. For now it's not too hard to keep the three in sync with correct data sources.

MrFlick:
You're right. More demos would probably be better. We just need to ensure that we keep them all in sync.

PR #56, opened 2016-05-20
Panel UI

Frencil:
# Overview

This branch expands panel-level user interface to include the following:

* The ability to define a panel title that appears in the top left corner of a panel
* A "panel controls" element that appears in the top right corner of a panel, including the following items:
  * A button to remove the panel
  * A button to display a panel description that can also be defined in the layout with arbitrary HTML
  * Buttons to move the panel up or down in terms of its y-index relative to adjacent panels
* All three panel control buttons are configurable in the layout, allowing for any combination of the three to be shown on a per-panel basis (default is to show all)
* Panel boundaries - draggable elements between panels that allow for adjusting the heights of panels one at a time

## Notes on the standard layout

Note that for the standard layout there is some dummy data in there for panel titles and descriptions that we'll likely want to remove or replace in consideration with how the Broad will consume the standard layout. Personally I think that such information is *very* implementation-specific and thus doesn't really belong in the standard layout at all.

Next week when we're working with Broad devs we can look at pulling things out of the standard layout into a Broad portal custom layout defined in their environment, or possibly look at spinning off a new layout (`BroadPortalLayout`?) that can live in the core library for now and allow the standard layout to be a bit more general.

## Feedback / Release outlook

While code review is always in season this branch is heavily about integration--how the final product feels to use. Is the new UI introduced on the branch intuitive? Clean? Is it obvious how it's supposed to work, and does it actually work that way? Where does it feel a bit clunky, or a bit loose?

It will be a subjective call as to when this branch is ready for master as the last 5% of any tight and snappy UI is a long tail that isn't necessarily in scope of this branch. It would be good to get this rolled into new release (0.3.9) by EOD Monday, May 23 in advance of the next integration session with LocusZoom's current primary consumer.

### DIFF
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -315,8 +315,8 @@ div.#{$namespace}-data_layer-tooltip-arrow_bottom_right {
 }
 
 .#{$namespace}-locuszoom-panel-controls {
-   padding-top: 4px;
-   padding-right: 4px;
+   padding-top: 6px;
+   padding-right: 2px;
 }
 
 div.#{$namespace}-panel-description {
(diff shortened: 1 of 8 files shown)


MrFlick:
This functionality looks good! If you can just clean up the conflicts, I'd be happy to merge.

Frencil:
Thanks! This was a fun branch. I've spotted a couple of areas I'd like to continue to refine but they're nuanced and it's time to keep moving. Conflicts should be resolved now.

PR #59, opened 2016-05-26
Tidy-up Data Sources

MrFlick:
This is a collection of smaller commits to tidy up the data source work. It's probably easier to look at each commit individually. The major revisions include:

1. Since `getURL` is not a required method for data.sources, the default caching behavior was modified to make sure it does return valid values before using it as the cacheKey
2. The `KnownDataSources` was converted from a simple global array to a proper singleton collection. This should make it easier to work with the default data sources
3. The `Data.Source.extend()` method now better allows for sub-classing existing data sources (preserving functions found on prototypes that might not be the base data source class)
4. The fields that the LD source uses for merging are now dynamically determined. This should allow it to merge with p-values from different namespaces without having to create multiple LD data sources
5. Some general linting, code cleanup, and increased test coverage


MrFlick:
The [wiki for Data Sources](https://github.com/statgen/locuszoom/wiki/Data-Sources) has been updated to show the more simple Data Source creation.

Frencil:
Looks great. The updates to LD in particular made it so in our example of adding panels/analyses (on `plot_builder.html`) we no longer had to define a new LD source for each new analysis added, so I updated the example to reflect that. Merging!

PR #60, opened 2016-05-31
Resolve Scalable Parameters

Frencil:
This pull req generalizes the resolution of scalable parameters (e.g. color, size, or shape for scatter points) such that every such layout parameter can be defined as:

* A static value (number or string independent of data)
* An object implementing a scale function to be applied to a field on the element's data
* An array of arbitrarily many of the above, to be iterated over until a non-null value is found or all options are exhausted.

There's also a new scale function in this branch called `if` that allows for a very basic conditional. Put it all together and we can support this definition for the color of scatter elements on the positions panel when the LD reference variant needs to be colored differently (e.g. we're coloring on the values of `ld:state` and `ld:isrefvar` with strict order):

```json
color: [
  {
    scale_function: "if",
    field: "ld:isrefvar",
    parameters: {
      field_value: 1,
      then: "#9632b8"
    }
  },
  {
    scale_function: "numerical_bin",
    field: "ld:state",
    parameters: {
      breaks: [0, 0.2, 0.4, 0.6, 0.8],
      values: ["#357ebd","#46b8da","#5cb85c","#eea236","#d43f3a"]
    }
  },
  "#B8B8B8"
]
```

The above layout, for each point on the scatter plot, is sent to `LocusZoom.DataLayer.resolveScalableParameter()` along with the data for that point. The first element in the array (the `if` scale function) will return its `then` value only if `ld:isrefvar` is 1, otherwise it will return null. This passes through to the next scale function, `numerical_bin`, which will map the value of `ld:state` to a breakpoint color or (if `ld:state` is not defined or numeric) will return null. That then passes to the last value, which is a string, and that gets returned no matter what. This ensures that LD ref variants are always purple, all other variants with a defined numerical LD value get an appropriate color, and any leftover variants show up gray.

The same behavior now applies to `point_size` and `point_shape` for scatter plot elements as well.

### DIFF
--- a/plot_builder.html
+++ b/plot_builder.html
@@ -94,7 +94,7 @@
     var apiBase = "http://portaldev.sph.umich.edu/api/v1/";
     var data_sources = new LocusZoom.DataSources()
       .add("base", ["AssociationLZ", {url:apiBase + "single/", params: { analysis: 3 }}])
-      .add("ld", ["LDLZ", "http://portaldev.sph.umich.edu/flask/v1/statistic/pair/LD/"])
+      .add("ld", ["LDLZ", apiBase + "pair/LD/"])
       .add("gene", ["GeneLZ", { url: apiBase + "annotation/genes/", params: { source: 2 } }])
       .add("recomb", ["RecombLZ", { url: apiBase + "annotation/recomb/results/", params: {source: 15} }])
       .add("sig", ["StaticJSON", [{ "x": 0, "y": 4.522 }, { "x": 2881033286, "y": 4.522 }] ]);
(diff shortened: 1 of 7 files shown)


plot_builder.html
```diff
@@ -94,7 +94,7 @@
     var apiBase = "http://portaldev.sph.umich.edu/api/v1/";
     var data_sources = new LocusZoom.DataSources()
       .add("base", ["AssociationLZ", {url:apiBase + "single/", params: { analysis: 3 }}])
-      .add("ld", ["LDLZ" ,apiBase + "pair/LD/"])
+      .add("ld", ["LDLZ", "http://portaldev.sph.umich.edu/flask/v1/statistic/pair/LD/"])
```
MrFlick: We probably shouldn't merge this particular URL into master. Other than that, this PR looks good.

PR #61, opened 2016-06-01
Exac gene constraints

Frencil:
This branch creates a known data source for hitting the ExAC gene constraint API and expands the genes panel and genes data layer tool tip to request and display gene constraint data.

Other core changes of note:

* `LocusZoom.createCORSPromise` now sets a `Content-Type` header when POSTing
* The main stylesheet now defines some default styles for headers and tables in tool tips
* `LocusZoom.parseFields` will now render an empty string if a field is present but the value is null (as opposed to doing nothing, leaving the field markup intact)

### DIFF
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -183,10 +183,12 @@ LocusZoom.Data.Source.prototype.getData = function(state, fields, outnames, tran
 };
 
 
-LocusZoom.Data.Source.prototype.parseResponse  = function(x, chain, fields, outnames, trans) {
-    var records = this.parseData(x.data || x, fields, outnames, trans);
-    var res = {header: chain.header || {}, body: records};
-    return res;
+LocusZoom.Data.Source.prototype.parseResponse = function(resp, chain, fields, outnames, trans) {
+    var json = resp;
+    if (typeof resp == "string"){ json = JSON.parse(resp); }
+    if (typeof json.data != "undefined"){ json = json.data; }
+    var records = this.parseData(json, fields, outnames, trans);
+    return {header: chain.header || {}, body: records};
 };
 
 LocusZoom.Data.Source.prototype.parseArraysToObjects = function(x, fields, outnames, trans) {
@@ -398,6 +400,7 @@ LocusZoom.Data.LDSource.prototype.getURL = function(state, chain, fields) {
 };
 
 LocusZoom.Data.LDSource.prototype.parseResponse = function(resp, chain, fields, outnames) {
+    var json = JSON.parse(resp);
     var keys = this.findMergeFields(chain);
     var reqFields = this.findRequestedFields(fields, outnames);
     if (!keys.position) {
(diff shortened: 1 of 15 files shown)


locuszoom.app.js
```diff
@@ -267,6 +267,11 @@ LocusZoom.createCORSPromise = function (method, url, body, timeout) {
         };
         timeout && setTimeout(response.reject, timeout);
         body = typeof body !== "undefined" ? body : "";
+        // If posting an object set the proper content-type header
+        if (method == "POST"){
+            xhr.setRequestHeader("Content-Type","application/x-www-form-urlencoded");
```
MrFlick: This wouldn't be the right content type for all POSTs. Any suggestions for how we should parameterize this? We could add another parameter to the function to pass in headers. Look to see if `body` is a special object with headers and content. Allow a `body` to be a function that we pass the `xhr` object to modify properties. Any other ideas?

It might also be a good time to take the default JSON parsing out of this function since some API endpoints might not necessarily return JSON. Maybe condition on the `responseType`

Frencil: Adding `headers` to the argument list seems to make the most sense since `body` is already there. I can get that in place pretty quickly and that would push the header definition into `LocusZoom.Data.GeneConstraintSource.prototype.fetchRequest`, which is a much more appropriate place for that definition to live (as it's also where we're specifying it's a POST).

As for removing the JSON parsing I agree, the response should just be passed through to the resolve method on the promise and the default parse data function should execute the JSON parsing, allowing any custom data source to do something different if need be.

MrFlick: Yeah, I think there is just a `response`, not sure why I used `resposeText` originally.

assets/js/app/Data.js
```diff
@@ -183,10 +183,12 @@ LocusZoom.Data.Source.prototype.getData = function(state, fields, outnames, tran
...
 
 
-LocusZoom.Data.Source.prototype.parseResponse  = function(x, chain, fields, outnames, trans) {
-    var records = this.parseData(x.data || x, fields, outnames, trans);
-    var res = {header: chain.header || {}, body: records};
-    return res;
+LocusZoom.Data.Source.prototype.parseResponse = function(resp, chain, fields, outnames, trans) {
+    var json = resp;
+    if (typeof resp == "string"){ json = JSON.parse(resp); }
+    if (typeof json.data != "undefined"){ json = json.data; }
```
Frencil: @MrFlick - I tried to make this, being the default `parseResponse` method, as generic as possible. The response coming back could already be parse JSON, hence the type check conditional, and not all responses will present all the data under a top-level `data` attribute, hence the check for that. It feels a bit messy, so let me know what you think.

MrFlick: Are there errors that this change is fixing? Are some of our responses not automatically being converted to JSON?

assets/js/app/Data.js
```diff
@@ -183,10 +183,10 @@ LocusZoom.Data.Source.prototype.getData = function(state, fields, outnames, tran
 };
 
 
-LocusZoom.Data.Source.prototype.parseResponse  = function(x, chain, fields, outnames, trans) {
-    var records = this.parseData(x.data || x, fields, outnames, trans);
-    var res = {header: chain.header || {}, body: records};
-    return res;
+LocusZoom.Data.Source.prototype.parseResponse = function(resp, chain, fields, outnames, trans) {
```
Frencil: @MrFlick - okay, take a look at this now. I refactored it once more to be a bit more in keeping with the original approach. The type checking is important because the StaticJSON data source uses the default parse method, and it's passing through an object, not a string representing an object (and `JSON.parse()` fails when fed an object as it casts its argument to a string, resulting in "[object Object]").

MrFlick: Ah. OK. I thought that the xmlhttprequest looked at the content type of the response but I see that it does not. Got it. Looks great.

PR #65, opened 2016-06-09
Fix `onUpdateFunctions` iteration issue

### LINKED ISSUE #64: Adding `onUpdate` function causes `undefined` to be added `onUpdateFunctions`
Observed this behavior today: I created a function `f() {}` (literally did nothing), then called `locusZoomPlot.onUpdate(f)`. That worked fine; however, the second call to `onUpdate()` (without argument, so therefore intending to run all of the stored functions) would yield a `Uncaught TypeError: this.onUpdateFunctions[func] is not a function`.

I replaced the `onUpdate` function with the following:
```js
LocusZoom.Instance.prototype.onUpdate = function(func){
    debugger;
    console.log('this called');
    console.log('func is', func, typeof func == 'undefined'); 
    console.log('currently', this.onUpdateFunctions);
    if (typeof func == "undefined" && this.onUpdateFunctions.length){
        console.log('these conditions met');
        for (func in this.onUpdateFunctions){
            console.log('func iterator', func);
            this.onUpdateFunctions[func]();
            console.log('in iterator', this.onUpdateFunctions);
        }
    } else if (typeof func == "function") {
        this.onUpdateFunctions.push(func);
    }
    console.log('now', this.onUpdateFunctions);
};
```

Using this, I could see that `f` got added correctly. However, when `onUpdate()` was called the first time, `this.onUpdateFunctions` became `[function function, undefined: undefined]`. Further inspection showed that when `onUpdateFunctions` was called the first time without argument, it would iterate over the following values: `0`, **`shuffle`, `swap`, `heapSort`**. `0` makes sense, as an array indice. The last 3 are surprising--`this.onUpdateFunctions["shuffle"]()` and `this.onUpdateFunctions["heapSort]()"` don't seem to do anything, but `this.onUpdateFunctions["swap"]()` seems to be adding the `undefined` to `onUpdateFunctions`. Not entirely sure where these are coming from--I assume it has something to do with the prototype methods for arrays in javascript, but some quick research didn't seem to show anything.

The fix seems to be swapping out the iteration like so:
```js
this.onUpdateFunctions.forEach(function(funcToRun) {
     funcToRun();
});
```

I'll make up a PR here in a minute.

tl;dr: the way JS handles iteration over an array doesn't always seem to work correctly, which results in `undefined` getting added to `onUpdateFunctions`

mssand-b:
Per #64, should fix the issue where adding functions to `onUpdateFunctions` causes `undefined` to be added to `onUpdateFunctions`

Fixes #64

### DIFF
--- a/locuszoom.app.js
+++ b/locuszoom.app.js
@@ -3035,9 +3035,9 @@ LocusZoom.Instance.prototype.rescaleSVG = function(){
 
 LocusZoom.Instance.prototype.onUpdate = function(func){
     if (typeof func == "undefined" && this.onUpdateFunctions.length){
-        for (func in this.onUpdateFunctions){
-            this.onUpdateFunctions[func]();
-        }
+        this.onUpdateFunctions.forEach(function(funcToRun) {
+            funcToRun();
+        });
     } else if (typeof func == "function") {
         this.onUpdateFunctions.push(func);
     }


Frencil:
Thanks for catching that, @mssand-b! I'm going to close this pull req, though, as the change needs to be made in the source that gets compiled down to `locuszoom.app.js` (see [ebb0858](https://github.com/statgen/locuszoom/commit/ebb0858c38a8eae10c1730100e1549a2fd453f7b) as well as [development setup documentation](https://github.com/statgen/locuszoom#development-setup)).

Since this was a very minor correction I just applied it on master. If you find yourself wanting to fix bugs you find in the core in the future, though, please take a look at the development setup instructions and if anything is ambiguous or unclear please let me know. It *should* be relatively quick to set up the proper node environment to run builds.

Issue #62, opened 2016-06-08
Provide methods for clearing selections


Frencil:
As seen in the conditional analysis proof of concept in the Broad portal implementation of LocusZoom, there may be situations where an implementer may want to programmatically clear selections. Presently this is possible but only through direct manipulation of the state, and important deselection logic therefore doesn't necessarily get triggered.

The LocusZoom core library should extend methods to clear all selected elements on a per-plot, per-panel, and per-data-layer basis.

Frencil:
This is addressed in PR #67

PR #67, opened 2016-06-15
Independent Select / Highlight / Tool Tip Logic

Frencil:
# Overview

This branch began when encountering a complication presented by conditional analysis on a variant: the variant disappears from view. It is still, however, tracked as a "selected" variant and therefore the selection must be cleared, making sure the tool tip goes with it.

Looking to provide functions to easily clear selections to an implementer I ended up organizing the code for highlighting, selecting, and showing tool tips for data elements. In the process I expanded the supported layout directives to allow more control over how highlighting, selecting, and tool tips should behave.

## New Methods

The `DataLayer` class has several new methods that will need to be documented for implementers to use when setting up custom interactivity:

* `highlightElement(element)`
* `unhighlightElement(element)`
* `highlightAllElements()`
* `unhighlightAllElements()`
* `selectElement(element)`
* `unselectElement(element)`
* `selectAllElements()`
* `unselectAllElements()`

The above should be more or less self explanatory. The `element` passed to the singular methods is a discrete element from the `data` array of the data layer.

## New layout syntax

The new layout syntax allows for defining select, highlight, and tool tip behavior independently:

```javascript
data_layer: {
    ...
    highlighted: {
        onmouseover: "on",
        onmouseout: "off"
    },
    selected: {
        onclick: "toggle_exclusive",
        onshiftclick: "toggle"
    },
    tooltip: {
        closable: true,
        show: { or: ["highlighted", "selected"] },
        hide: { and: ["unhighlighted", "unselected"] },
        html: ""
    }
```

In this example, a mouse over event will highlight an element and a mouse out event will unhighlight it. A click will toggle the selection of an element exclusively (if selected deselect it, if not selected select it and deselect all the other elements). A shift click will just toggle with no exclusivity.

The tool tip also has syntax to dictate when it should be shown or hidden based on the selected and highlighted statuses together. Combined these new layout directive provide independent definitions for selection and highlighting, and tool tip display logic that is explicitly coupled to those statuses by the implementer to the behavior of their choosing.

In the process of setting this up it was little extra effort to include support for control-clicking, shift-clicking, and control-shift-clicking so that is included too.

### DIFF
--- a/assets/css/locuszoom.scss
+++ b/assets/css/locuszoom.scss
@@ -107,7 +107,7 @@ svg.#{$namespace}-locuszoom {
     cursor: pointer;
   }
 
-  path.#{$namespace}-data_layer-scatter-hovered {
+  path.#{$namespace}-data_layer-scatter-highlighted {
     stroke: #{$default_black_shadow};
     stroke-width: 4;
   }
@@ -153,7 +153,7 @@ svg.#{$namespace}-locuszoom {
     stroke-width: 0;
   }
 
-  rect.#{$namespace}-data_layer-genes.#{$namespace}-data_layer-genes-bounding_box-hovered {
+  rect.#{$namespace}-data_layer-genes.#{$namespace}-data_layer-genes-bounding_box-highlighted {
     fill: rgb(54, 54, 150);
     fill-opacity: 0.1;
     stroke-width: 0;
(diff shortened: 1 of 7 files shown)


Frencil:
This addresses Issue #62

Issue #68, opened 2016-06-16
Structure panels and data layers as arrays in the layout


Frencil:
As described by @jonathonl, the current convention for defining lists of primary objects in the layout (panels and data layers) is confusing:

```
var layout = {
  ...
  panels: {
    panel_1_id: {
      ...
      data_layers: {
        data_layer_1_id: {
          ...
        }
      }
    },
    panel_2_id: {
      ...
    }
  }
};
```

The problem here is that panel and data layer IDs are expressed as keys in a map, a space generally reserved for supported/documented attribute names, not user-defined values. A more intuitive approach would make the id value a named parameter and use arrays, like so:

```
var layout = {
  ...
  panels: [
    {
      id: "panel_1_id",
      data_layers: [
        {
          id: "data_layer_1_id",
          ...
        }
      ]
    },
    {
      id: "panel_2_id",
      ...
    }
  ]
};
```

While this is a pretty significant breaking change it makes the layout much more internally consistent.

Frencil:
Resolved in v0.4.0

PR #69, opened 2016-06-17
Layout array syntax

Frencil:
This branch refactors the layout syntax for panels and data layers to be arrays of objects. This addresses Issue #68.

Essentially where before we had this:

```
var layout = {
  ...
  panels: {
    panel_1_id: {
      ...
      data_layers: {
        data_layer_1_id: {
          ...
        }
      }
    },
    panel_2_id: {
      ...
    }
  }
};
```

We now have something like this:

```
var layout = {
  ...
  panels: [
    {
      id: "panel_1_id",
      data_layers: [
        {
          id: "data_layer_1_id",
          ...
        }
      ]
    },
    {
      id: "panel_2_id",
      ...
    }
  ]
};
```

This is more internally consistent as with this change all customizable data (i.e. panel IDs and data layer IDs, in addition to all the other layout "values") are consistently to the right side of the colon in any place in the layout. The left side of the colon anywhere in the layout is now, consistently, a documented supported property of the layout.

Methods for adding panels and data layers got slightly simpler as they now just take a layout and do sanity checking to make sure an ID is present in the layout (before they took ID and layout separately).


MrFlick:
Looks good. Do we really need to require an ID? Can we assign some random name if they don't supply one? If naming things is hard, it might be nice not to require it if it's not necessary.

Frencil:
> Do we really need to require an ID? Can we assign some random name if they don't supply one? If naming things is hard, it might be nice not to require it if it's not necessary.

Good call. I just pushed an update that will generate an ID if not present in the layout (with tests) and will still throw an error if passed a panel layout that explicitly specifies a panel ID that's already in use (also with a test). ID generation is of the form "p" followed by a pseudorandom 8 digit number.

Issue #17, opened 2016-01-18
Mousewheel zooming on data layers


Frencil:
** Depends on #15 and #16**

When a user mouses over a data layer and scrolls with their mouse wheel the data layers should "zoom" by changing the x range of the visible data set.

Based on a previous discussion with Broad developers the desired default behavior would be to scale the y range on the positions panel to fit all data regardless of the previous view's y range.

## Estimate

This feature will probably take two weeks of development to build and test.

MrFlick:
This should probably also include "pinching" on mobile devices.

Issue #16, opened 2016-01-18
Drag panels horizontally to remap to new data (panning)


Frencil:
**Depends on #15**

When a user clicks and drags horizontally on a data layer's background the charts and axes should drag left/right, and when completing a drag event should trigger a remap event to render new data based on new regions now in view.

Based on a previous discussion with Broad developers the desired default behavior would be to scale the y range on the positions panel to fit all data regardless of the previous view's y range.

## Estimate

This feature will probably take two weeks of development to build and test.

PR #76, opened 2016-09-09
Panel Interaction

Frencil:
# Overview

This branch introduces several mouse interaction features. These include:

* Scroll the mousewheel over a panel to zoom relative to the x axis
* Click and drag a panel's background to pan left/right along the x axis
* Click and drag an x axis tick to scale the x axis
* Shift-click and drag an x axis tick to pan left/right along the x axis
* Click and drag a y axis tick to scale the y axis (each y axis behaving independently)
* Shift-click and drag a y axis tick to pan up/down along the y axis (each y axis behaving independently)

## New layout directives

All interactions are configurable in panel layouts. Here's an overview of the new supported directives:

* **`panel.interaction.drag_background_to_pan`** - boolean; whether to allow background panning
* **`panel.interaction.drag_x_ticks_to_scale`** - boolean; whether to allow dragging x ticks to pan/scale x axis
* **`panel.interaction.drag_y1_ticks_to_scale`** - boolean; whether to allow dragging y1 ticks to pan/scale y1 axis
* **`panel.interaction.drag_y2_ticks_to_scale`** - boolean; whether to allow dragging y2 ticks to pan/scale y2 axis
* **`panel.interaction.scroll_to_zoom`** - boolean; whether to allow scrolling to zoom along x axis (zooming either y axis not supported at this time)
* **`panel.interaction.x_linked`** - boolean; whether interactive changes to the x axis are linked to other panels
* **`panel.interaction.y1_linked`** - boolean; whether interactive changes to the y1 axis are linked to other panels
* **`panel.interaction.y2_linked`** - boolean; whether interactive changes to the y2 axis are linked to other panels

All of the above default to false in the default panel layout. The standard layout, used by many LocusZoom implementations, enables all interactive booleans and links the x axes on the two panels (association data and genes) that appear in that layout.

## Panel linking and interaction broadcasting

The problem to solve is that in some cases two or more panels may be "linked" on a given axis; for example, if the x axis in an association panel changes due to interaction of any kind then the x axis in a genes panel should also update in real time, not just when new data is loaded. The layout directives `x_linked`, `y1_linked`, and `y2_linked` are a simple approach to this problem.

Essentially, if an interaction is being performed on a panel (e.g. actively dragging the background to pan) a check is performed continuously. The check first determines which axis is being changed by the interaction, and then looks to see if the panel in question has that axis linked (by way of the layout directive). If so, it then queries all other panels to see if they are also linked on that same axis, and if so, broadcasts the objects storing the interaction data over to that panel. In this way the linked panel has its own copy of the interaction information, as if the user were dragging all linked panels in the same way at the same time.

Where this simple approach may fall short is if a single LocusZoom plot would have the need to link panels in multiple groups for the same axis (e.g. have four panels, link panels 1 and 2 on their x axes, link panels 3 and 4 on their x axes). It seems at this point that this level of abstraction may not be necessary, as spinning off a second group of panels into a distinct LocusZoom plot may be an effective workaround.

## Issues covered

This branch addresses Issues #16 and #17. With the addition of y-axis scaling and panning it also addresses the as-yet-undocumented issue of needing to adjust y axes to focus on a particular range.

This branch does not yet address the proposed option of introducing "breaks" in a y-scale (for example, on a -log10 pvalue scale, showing 0-10 with a break and then 6-70). My hope is that the ability to effectively "zoom in" on any continuous range on either y-axis will mitigate the immediate need for this feature, though it seems we'll still need that eventually too.


Frencil:
Reverted to active development... feature complete but missing tests. Will commit tests shortly.

Frencil:
Tests added, branch now truly development complete. Ready for testing/review/feedback.

PR #91, opened 2017-03-12
Implement conditional blocks in parseField templates

pjvandehaar:
I haven't tested anything but latest Chrome.

The test is for this template:

```
{{#if foo:field_1}}
  <strong>
    {{foo:field_1}}
    {{#if bar:field2}}
       and {{bar:field2}}
    {{/if}}, 
    {{#if nope}}
      wat
    {{/if}}
    {{bar:field2|herp|derp}}; 
    {{field3}}
  </strong>
{{/if}}
```

It runs iteratively, always handling the last if-then block in the template, in order to handle nesting correctlyish.  Unfortunately, that means that users can make templates like: 
```
{{#if pvalue{{#if foo}}|neglog10{{/if}}}}
  {{pvalue{{#if foo}}|neglog10{{/if}}}}
{{/if}}
```

There is no way for user data to meaningfully affect program execution, so it seems fine.

I don't know what happens when errors occur.  I don't see any way to make an infinite loop.

Mustache and Handlebars parse an AST and then evaluate that.  Good for them.

### DIFF
--- a/assets/js/app/LocusZoom.js
+++ b/assets/js/app/LocusZoom.js
@@ -319,6 +319,25 @@ LocusZoom.parseFields = function (data, html) {
     if (typeof html != "string"){
         throw ("LocusZoom.parseFields invalid arguments: html is not a string");
     }
+    // Handle conditional blocks {{#if attr}}...{{/if}} from back to front to handle nesting correctly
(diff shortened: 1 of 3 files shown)


Frencil:
Looking at this tonight; thanks for taking a crack at it! I may expand the tests a bit before merging to cover more cases, and hopefully some corner cases.

PR #92, opened 2017-03-13
Use an intermediary AST in LZ.parseFields()

pjvandehaar:
This avoids some of the odd behavior of the previous conditional-template implementation.  If the AST were cached it could also be much faster, but oh well.

### DIFF
--- a/assets/js/app/LocusZoom.js
+++ b/assets/js/app/LocusZoom.js
@@ -355,12 +355,7 @@ LocusZoom.parseFields = function (data, html) {
 
     var resolve = function(variable) {
         if (!resolve.cache.hasOwnProperty(variable)) {
-            try {
-                resolve.cache[variable] = (new LocusZoom.Data.Field(variable)).resolve(data);
-            } catch (error) {
-                console.error('Error while evaluating field ' + JSON.stringify(variable) + ': ' + error);
-                resolve.cache[variable] = variable;
-            }
+            resolve.cache[variable] = (new LocusZoom.Data.Field(variable)).resolve(data);
         }
         return resolve.cache[variable];
     };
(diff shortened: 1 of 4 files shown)


assets/js/app/LocusZoom.js
```diff
@@ -319,42 +319,54 @@ LocusZoom.parseFields = function (data, html) {
...
+            resolve.cache[variable] = (new LocusZoom.Data.Field(variable)).resolve(data);
+        return resolve.cache[variable];
+    };
+    resolve.cache = {};
+    var render_node = function(node) {
+        if (node.text) {
+            return node.text;
+        } else if (node.variable) {
+            var value = resolve(node.variable);
+            return (["string","number","boolean"].indexOf(typeof value) != -1) ? value :
```
Frencil: I'd recommend splitting this out into a more verbose conditional. A ternary operator inside another ternary operator reduces clarity and makes for difficult long-term maintenance

assets/js/app/LocusZoom.js
```diff
@@ -319,42 +319,54 @@ LocusZoom.parseFields = function (data, html) {
...
+    var tokens = [];
+    var regex = /\{\{(?:(#if )?([A-Za-z0-9_:\|]+)|(\/if))\}\}/;
+    while (html.length > 0){
+        var m = regex.exec(html);
+        if (!m) { tokens.push({text: html}); html = ''; }
+        else if (m.index != 0) { tokens.push({text: html.slice(0, m.index)}); html = html.slice(m.index); }
+        else if (m[1] == "#if ") { tokens.push({condition: m[2]}); html = html.slice(m[0].length); }
+        else if (m[2]) { tokens.push({variable: m[2]}); html = html.slice(m[0].length); }
+        else if (m[3] == "/if") { tokens.push({close: 'if'}); html = html.slice(m[0].length); }
+        else { throw [m, html] }
```
Frencil: I thought it was probably okay to have `throw` in `parseFields` in #91 but I actually think it'd be better to wrap this while loop in a try/catch and handle any thrown exceptions within the method. The reason here is that it's not the core library that's failing, we're just trying to parse some badly marked up HTML.

At the very least we should catch any such exceptions and pipe them gracefully to the console, then try our best to parse and return the HTML we've been given with any bad syntax untouched. This way `parseFields` can be trusted to always return a string, so in the case of tool tips, a tool tip will always render (but someone trying to make LZ do something clever will see unparsed stuff in their tool tips with detail in the console instead of an uncaught exception).

assets/js/app/LocusZoom.js
```diff
@@ -319,42 +319,54 @@ LocusZoom.parseFields = function (data, html) {
...
+    var astify = function() {
+        var token = tokens.shift();
+        if (token.text || token.variable) {
+            return token;
+        } else if (token.condition) {
+            token.then = [];
+            while(tokens[0].close != 'if') token.then.push(astify());
+            tokens.shift(); //consume {{/if}}
+            return token;
+        } else { throw 'what?' }
```
Frencil: Same as other comment on throws... let's handle this a bit more gracefully.

assets/js/app/LocusZoom.js
```diff
@@ -319,42 +319,54 @@ LocusZoom.parseFields = function (data, html) {
...
+        if (node.text) {
+            return node.text;
+        } else if (node.variable) {
+            var value = resolve(node.variable);
+            return (["string","number","boolean"].indexOf(typeof value) != -1) ? value :
+                (value == null) ? '' :
+                '{{' + node.variable + '}}';
+        } else if (node.condition) {
+            return resolve(node.condition) ? node.then.map(render_node).join('') : '';
+        } else { throw 'foo' }
```
Frencil: Same as other comment on throws... let's handle this a bit more gracefully.

assets/js/app/LocusZoom.js
pjvandehaar: I've always loved the way that order-of-operations lets you write switch-case-like nested ternaries.  Anyways, I've changed it.

Frencil:
Looks good. FYI all library JS and tests are linted using the rule set [here](https://github.com/statgen/locuszoom/blob/master/.eslintrc) (see latest commit on branch where I cleaned up some linting issues). Something to keep in mind for future dev work.

PR #94, opened 2017-03-13
Let tooltip content flow around close button (x)

pjvandehaar:
This is gross and I will improve it but it works well.


Frencil:
If you're going to go down this road just make sure to take a look at the `lz-tooltip-close-button` class (in locuszoom.scss) so we're not defining classes that get immediately overridden with inline styling.

And again, warning about arbitrary HTML content... try to feed some HTML to a tool tip with a close button done up this way that has some arbitrary floated elements. Can you make it clash-proof?

Frencil:
This one should be ready now - needed relative positioning and margins

PR #97, opened 2017-03-14
Make mouse_guide optional and disable it in standard_phewas plot

pjvandehaar:


pjvandehaar:
This second commit makes changing `layout.mouse_guide` at runtime have no effect.  I don't know whether that's good or bad.  If someone had two LZ instances, one named `foo` and the other named `foo-mouse-guide`, I don't know whether they will have problems.  Probably not, because they're not registering their events to the same DOM elements, right?  Why did you add namespaces in the first place?

Frencil:
The namespaces were probably not necessary since they're bound to the svg... they were probably originally bound to the page in a much earlier version. They're doing no harm, though. At any rate, I just made the `mouse_guide` directive true in the default layout for all plots (defined in Plot.js) - this is important as it ensures that value is always defined, which having the added benefit of only needing to explicitly set it when it should be false. With that this branch is ready to go!

pjvandehaar:
Great, thanks.

PR #102, opened 2017-05-23
Include all keys sent from DataSource even if not listed in `fields`

pjvandehaar:
(only implemented for ObjectOfArrays APIs, but I'll do the other half if you think this is fine.)

@Frencil said he'd like something like this. @MrFlick , does this seem fine to you?

If `fields` contains `["id", "chr", "position", "pvalue|neglog10"]`, should we also include `pvalue` (untransformed) in the output of the DataSource?  If the layout explicitly listed `pvalue|neglog10` in `.fields`, maybe it won't use `pvalue` or any other transform of it?

### DIFF
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -235,7 +235,17 @@ LocusZoom.Data.Source.prototype.parseArraysToObjects = function(x, fields, outna
     fields.forEach(function(f, i) {
         if (!(f in x)) {throw "field " + f + " not found in response for " + outnames[i];}
     });
-    var N = x[Object.keys(x)[1]].length;
+    var x_keys = Object.keys(x);
+    var N = x[x_keys[0]].length; // NOTE: this used 1 instead of 0 before, why?
+    x_keys.forEach(function(key) {
+        if (x[key].length !== N) {
+            throw "the response column " + key + " had " + x[key].length.toString() +
+                " elements but " + x_keys[0] + " had " + N.toString();
+        }
+    });
+    var nonfield_keys = x_keys.filter(function(key) {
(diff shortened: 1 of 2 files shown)


pjvandehaar:
(tests are passing on my machines, I still don't understand what's wrong with Travis)

assets/js/app/Data.js
```diff
@@ -235,7 +235,17 @@ LocusZoom.Data.Source.prototype.parseArraysToObjects = function(x, fields, outna
     fields.forEach(function(f, i) {
         if (!(f in x)) {throw "field " + f + " not found in response for " + outnames[i];}
     });
-    var N = x[Object.keys(x)[1]].length;
+    var x_keys = Object.keys(x);
+    var N = x[x_keys[0]].length; // NOTE: this used 1 instead of 0 before, why?
```
Frencil: Looking at the type of responses that come back where this method is invoked using 1 instead of 0 seems arbitrary. It could be that the endpoint originally gave back an object where the first key was not an array of length consistent with the other keys, but that no longer appears to be the case. It's certainly less safe to assume the object coming back will have at least two keys, so I think you're in the right to look at the 0th key here and can safely drop this comment.

abought:
#134 provides an alternative, via overriding `extractFields`. (which a subclass, like the genes source, would need to opt in to doing) This change will land in v0.8, targeted for the end of this month.

With that hook, is this more global change still a priority? If so, we can keep this open to revisit. Otherwise, I'll close this ticket when 0.8 comes out.

abought:
Closing this ticket as the alternate was changed, and there's been no further discussion. If we find a pheweb use case where this is highly useful, I'd be happy to revisit at that later date.

PR #108, opened 2017-08-04
Add ESLint to default test command

abought:
## Purpose
Add ESLint to the default test command, to help catch common errors. In particular, I've updated the rules and code to [de-emphasize the use of magic type coercion](https://www.destroyallsoftware.com/talks/wat) (`== in favor of ===`), which should help prevent edge case bugs as we explore more custom plots in the future. Any tests that relied on this behavior have been fixed.

Also added an npm5 lockfile to the repository, to make dependencies more deterministic. ([details](https://docs.npmjs.com/files/package-locks))

Lastly, updated travis to use caching, which should enable faster builds. Travis will now require that all linting pass as part of any pull request.

## Tips
- If you're not used to the style guide, `eslint --fix .` will automatically reformat for simple conventions (like indents), to make usage less burdensome.
- If you don't have eslint installed globally, then `npm run eslint` will save you from having to type the full path to the executable.
- Use `npm test` to subject your code to the exact rules that Travis uses. I have omitted eslint from the gulp task intentionally, to keep the `gulp watch` workflow fluid.

## Testing notes
All mocha tests passing. As this is a first PR to this repo, I'll do a pass of manual testing on Monday, and then may recruit someone with custom LZ plots for manual verification.

(Custom or third party plots are probably a bit more likely to test the limits of what plots do, or to use the code in novel or unique ways)

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,14 @@
+# Build-related assets
+assets/js/vendor
+assets/js/app/wrapper.js
+# Non-code items
+docs/
+examples/
+node_modules/
+staticdata/
+package.json
+
+# Built files
+locuszoom.app.js
+locuszoom.app.min.js
+locuszoom.vendor.min.js
--- a/.eslintrc
+++ b/.eslintrc
@@ -2,7 +2,11 @@
     "rules": {
         "indent": [
             2,
-            4
+            4,
+            {
+                "FunctionExpression": { "parameters": "first" },
+                "CallExpression": { "arguments": "first" }
+            }
         ],
         "quotes": [
             2,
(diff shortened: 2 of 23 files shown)


abought:
Ran through manual regression testing on all current sample plots, and working as expected. [1] Will go ahead and merge, but not cut a release for this.

@pjvandehaar , these are the tiny changes we discussed last week. I don't expect it to affect your work, but let me know if PheWeb has trouble.

[1] I did see some issues with p-value tooltips not appearing in the PheWAS forest plot, but it also occurs on master; fix is out of scope for this PR.
 https://github.com/statgen/locuszoom/blob/master/examples/phewas_forest.png

PR #110, opened 2017-08-29
JSDoc source code documentation

abought:
⛷ Please squash during merge

# Purpose
Add source code docstrings in the JSDoc format. There is a lot of good information about methods in the wiki, and this PR adds those descriptions directly to the code.  I am also adding extra type annotations for methods.

With future improvements, this could be used to  auto-generate cross-linked HTML docs that would complement the wiki prose/tutorials.

# TODO
I've been annotating code as I work through the codebase (to capture my own notes). Some of these docstrings are tentative and will be firmed up as I go along (eg public/private APIs, method types). I'll keep adding docs as time/tasks allow.

# Review note
I recommend merge-and-squash for this PR; the commit history is a reflection of process rather than need for separate commits.


assets/js/app/DataLayers/genes.js
```diff
@@ -34,18 +41,40 @@ LocusZoom.DataLayers.add("genes", function(layout){
...
+     * A gene may have arbitrarily many transcripts, but this data layer isn't set up to render them yet.
+     * Stash a transcript_idx to point to the first transcript and use that for all transcript refs.
+     * @member {number}
+     * @type {number}
+     */
     this.transcript_idx = 0;
-    
+
+    /**
+     * TODO: Description. How does this differ from similarly named properties?
```
Frencil: `this.tracks` is nothing more than a number that stores the number of tracks in the data layer. Each time a new region is requested an arbitrary number of genes will come back with arbitrary overlaps, so `assignTracks()` runs. It resets `this.tracks` to 1 and cycles through, positioning genes on different tracks to prevent overlap while (ideally) keeping tracks to a minimum.

Tracks on a genes panel are not special or indicative of any meaning, they're just rows used to vertically separate overlapping genes.

assets/js/app/DataLayers/genes.js
```diff
@@ -34,18 +41,40 @@ LocusZoom.DataLayers.add("genes", function(layout){
...
+     * After we've loaded the genes interpret them to assign
+     *   each to a track so that they do not overlap in the view
+     * @returns {LocusZoom.DataLayer}
+     */
     this.assignTracks = function(){
-
-        // Function to get the width in pixels of a label given the text and layout attributes
+        /**
+         * Function to get the width in pixels of a label given the text and layout attributes
+         *      TODO: Move to outer scope?
```
Frencil: Yeah, could be useful outside of track assignment some day.

assets/js/app/Layouts.js
```diff
@@ -4,6 +4,14 @@
 
 "use strict";
 
+/**
+ * Manage known layouts for all parts of the LocusZoom plot
+ *
+ * This registry allows for layouts to be reused and customized many times on a page, using a common base pattern.
+ *   It handles the work of ensuring that each new instance of the layout has no shared state with other copies.
```
Frencil: :+1:

assets/js/app/Panel.js
```diff
@@ -5,21 +5,21 @@
...
 
 /**
-
-  LocusZoom.Panel Class
-
-  A panel is an abstract class representing a subdivision of the LocusZoom stage
-  to display a distinct data representation
-
+ * A panel is an abstract class representing a subdivision of the LocusZoom stage
+ *   to display a distinct data representation
```
Frencil: Might be worth expanding a bit:

> *   to display a distinct data representation as a collection of data layers

assets/js/app/Panel.js
```diff
@@ -308,7 +431,12 @@ LocusZoom.Panel.prototype.setTitle = function(title){
     return this;
 };
 
-// Initialize a panel
+
+/**
+ * Initialize the panel
+ * TODO: Clarify uninitialized panel (eg data vs rendering?)
```
Frencil: Initialize is essentially the panel's first draw function, in that a panel is just a stage for data layers and their scaffolding (e.g. axis if applicable, the curtain/loader elements, the actual `<g>` tag inside which all the child data layers do their rendering, etc.). Panels also have a `render` function, but that's the "loop" that gets called when anything triggers a re-draw (it updates the already-initialized scaffolding display elements and triggers a redraw of child data layers if needed, among other things).

An uninitialized panel may have its *layout* initialized (by `initializeLayout`... overloaded term to be sure) which means the layout "agrees with itself" (e.g. has sane dimensions / origin / margins with respect to discrete and proportional values, as well has having datalayers defined in the layout instantiated as children.

assets/js/app/Panel.js
```diff
@@ -527,15 +686,22 @@ LocusZoom.Panel.prototype.removeDataLayer = function(id){
...
     this.data_layer_ids_by_z_index.forEach(function(id){
         this.data_layers[id].setAllElementStatus("selected", false);
     }.bind(this));
     return this;
 };
 
-// Re-Map a panel to new positions according to the parent plot's state
+/**
+ * When the parent plot changes state, adjust the panel accordingly. For example, this may include fetching new data
+ *   from the API as gene position changes
```
Frencil: Again position isn't the best term and "gene position" can be more misleading as not all LZ configurations need to include a genes data layer. Better to use region.

> For example, this may include fetching new data from the API as region is changed.

assets/js/app/DataLayer.js
```diff
@@ -109,6 +144,10 @@ LocusZoom.DataLayer.prototype.getElementId = function(element){
     return (this.getBaseId() + "-" + element_id).replace(/(:|\.|\[|\]|,)/g, "_");
 };
 
+/**
+ * @param {String} id
```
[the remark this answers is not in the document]
abought: Could I ask you to double-check this description?

This method returns `selector.data()[0]` rather than `selector`, and my reading would actually be as follows:

> Returns a reference to the underlying data associated with a single visual element in the data layer, as referenced by the unique identifier for the element

This method isn't used terribly often, but it seems like this description would be more in the spirit of how this method is used* as well, unless I'm missing an edge case.

If the new description is ok, I might even propose adding this method to the list of variables/methods to be renamed. I'll usually try to keep code changes to a separate commit, so that anything that might introduce a bug can be cleanly reverted.

* For example, the return value of this method is eventually passed (as `d`) to `updateTooltip`, which explicitly uses it in data binding:
https://github.com/abought/locuszoom/blob/d0e12ffd4f7727dd4998fa98de5dff16cccda432/assets/js/app/DataLayer.js#L403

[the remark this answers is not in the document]
Frencil: Good catch - I hadn't re-read the function that carefully and went off fallible memory. Your proposed description is best.

As for renaming - I'm all for adding clarity, and admittedly I was thrown off by this method's name coming back with a few months' fresh eyes. It looks like this method is only currently used for selecting elements (meaning here members of a data set inside a data layer) that have a named status (e.g. "highlighted", "hidden", etc.) explicitly when needing to toggle that status *off*.

See [DataLayer.setAllElementStatus()](https://github.com/abought/locuszoom/blob/d0e12ffd4f7727dd4998fa98de5dff16cccda432/assets/js/app/DataLayer.js#L758-L772) one of the only places `getElementById()` is called. If we're toggling a status *on* for the whole data layer (e.g. highlight all) then we can just loop through all elements in the `data` array. We might touch some elements that are already highlighted / selected / whatever, but it's probably more likely that only a few are in this status (since the default for every status is off) so it's probably not worth filtering the list. A selected element will still be selected since we're being explicit about what the value of the status should be.

To toggle *off*, though, it makes less sense to loop through every element. I could, for example, select a single element of a populous data set (say 1000+ pvalues on an association data layer) and then click the background to trigger a "deselect all". Suddenly I'm looping through 1000 elements? I can't remember if I ever saw a performance drag here, but for such a common action the optimization of looking up elements with the status *on* (in the `state` object where status / element ID maps are stored) would almost always be faster, and that warranted the creation of a means to perform that lookup - `getElementById`.

Looking at this after not working with it for a while I'm surprised to see this is still the only real use case of this method, but that's the thought process that underpins it.

A good way to see this in action is by inspecting a plot. As you work with LocusZoom I hope you find live plots easy to deconstruct and explore in the console - keeping the plot object structure clean while remaining fully representative of a plot was a major design pattern that I tried to apply throughout...

1. Go to [statgen.github.io/locuszoom/](http://statgen.github.io/locuszoom/) and open the console
2. The example plot is stored in the global variable `plot`. Inspecting that alone in the console puts you at the root of the plot object on the page.
3. Click the gene TCF7L2 to select it.
4. In the console inspect `plot.panels.genes.data_layers.genes` (the genes data layer inside the genes panel)
5. Inspect the `data` object (`plot.panels.genes.data_layers.genes.data`) and see its an array containing objects representing the genes seen on the page (TCF7L2 is probably index 3)
6. Inspect the `state` object at any level of the plot (off the plot itself or any panel or data layer). They're all references to the same object in memory. Attributes in the state object are organized underneath keys representing their depth. Region-related attributes like `chr` / `start` / `end` are plot-wide so top-level. Other keys represent all panels and data layers by their dot-concatenated IDs.
7. `genes.genes` is the object in `state` for the genes data layer in the genes panel (nothing logically wrong about the ID being duplicated but for the main layout it could be better). In there are arrays for each status, and all should be empty except for `selected` that should have one ID string. That string should be present on one of the elements in `plot.panels.genes.data_layers.genes.data` and its key should be whatever is defined as the `id_field` on the data layer in the layout (`plot.panels.genes.data_layers.genes.layout.id_field` - which in this example is `gene_id` and drives home why that pattern is there - LocusZoom is agnostic to what part of an element in a collection from an arbitrary data source is the ID... *you* specify what that is in the layout).

**

A lot behind one little function, but a good way to see how some of these pieces currently fit together.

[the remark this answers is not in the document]
abought: Just a small note- the methods for introspecting plots have been both welcomed and appreciated; in some cases I've been using them to help identify internal structure for the docs.

If we keep getting questions about implementing LZ on custom servers, it's a big plus to know that the essentials of a good debugging guide are already in place! 🐙 

(the genes layer is a bit of a rebel in some ways, but that's not an 11pm thing!)

I've had two notes recently that expressed interest in contributing back to core LZ, so I'll try to chug along on some more of docs this weekend and capture the insights you've been providing.

abought:
Did a final quick pass through and merged docs as they are.

There may be improvements warranted in the future, but I've merged as is to make room for future improvements.

PR #111, opened 2017-09-12
[hold] Improve user-facing error message

abought:
# Purpose
Patch proposed by @benralexander . PR created on his behalf by request.

A developer-focused error was being shown to users (likely in [this block](https://github.com/abought/locuszoom/blob/24c5a3780914b0bcd80ca9fd9d0797e2bb9e3f0a/assets/js/app/Panel.js#L543-L563)) Replace that error with more user-friendly text.

Ben provides the following notes about how to reproduce:
> If users sometimes specify a genomic range in which no variants exist (definitely easy to do in some of our data sets, and especially if they have 'zoomed in') then the default error message was leaving them confused.  The message we've provided instead at least gives them a hint about what they should do next.  Clearly a more general approach would provide the calling routine with more options for error handling in this common case.

### DIFF
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -430,7 +430,8 @@ LocusZoom.Data.LDSource.prototype.getURL = function(state, chain, fields) {
             var columns = "";
             if (!keys.id){ columns += (columns.length ? ", " : "") + "id"; }
             if (!keys.pvalue){ columns += (columns.length ? ", " : "") + "pvalue"; }
-            throw("Unable to find necessary column(s) for merge: " + columns + " (available: " + keys._names_ + ")");
+            console.error("Unable to find necessary column(s) for merge: " + columns + " (available: " + keys._names_ + ")");
+            throw("No variants in range for this data set.  Expand the range, or try a different data set");
         }
         refVar = chain.body[findExtremeValue(chain.body, keys.pvalue)][keys.id];
     }


pjvandehaar:
Is it possible for this error to occur when variants _were_ returned but lacked the fields `id` and `pvalue`?

abought:
Peter and I traced through some logic earlier, and I concur. Will talk to Ben about this proposal during tomorrow's technical call.

Although a developer facing error message shouldn't be shown to users, I like that there is a better option for suppressing it upstream in the code. 

Long term, I think this exposes a challenge for new users, in that the data sources sometimes embody complex and API-specific assumptions about data sources and required payload structures. (Given that data sources are swappable, we're still doing pretty well for modularity overall!)

We'll see what Ben says, and I can help them trace the difference in environments further if needed.

benralexander:
That all sounds good to me, and I can implement the strategy that Chris
describes.  Clearly modifying the core library with an error message that
addresses only the symptom our users were seeing was only a work-around,
and I'm happy to have a more systematic solution available.

Thanks for the help!

abought:
Thanks Ben! We'll touch base later today, and can close the PR once we're sure the proposed alternative works for you. I'll also target adding some documentation around the issue uncovered here.

abought:
Brief follow up: @benralexander , have you had an opportunity to try the alternate proposed fix yet? 

Current plan is to merge / test a few things at once this Wednesday; currently this PR serves as a tracking ticket, and may be closed at that time to avoid developing a backlog.

abought:
Ben has provided a reproducible test case on the portal (a region where we find no available data, without needing to scroll too far). I'll log here for now, and resume debugging as time allows. (hopefully next week)

1. Go to the ["metabochip" dataset](http://www.type2diabetesgenetics.org/variantInfo/variantInfo/8_118185025_G_A) (which has clusters of data) and select the Locuszoom plot option

2. Zoom in (eg to a narrow band such as 90kb), using shift+scroll if necessary (there may not be zoom buttons). Pan left so that your screen is centered around a region without data; this one worked for me: `Region: 8:118105025-118265025`

3. The plot should show Ben's custom error message, which modifies LZ source code.
screen shot 2018-01-11 at 9 30 48 am

Here is where the error message is implemented in their fork, for comparison:
https://github.com/broadinstitute/dig-diabetes-portal/blob/master/web-app/js/lib/locuszoom.app.js#L7801

abought:
One more small trick: rather than scrolling manually, this can be controlled from the developer JS console in the browser once the linked dataset is loaded. For my own future notes, here are the commands used to reach a region with no GWAS data. (worked as of Jan 11, 2018)

```javascript
var thePlot = mpgSoftware.locusZoom.locusZoomPlot['#lz-47'];
thePlot.state.start = 118081387;
thePlot.state.end = 118171499;
thePlot.applyState();
```

abought:
Per @benralexander meeting: we'd like to make the error message customizable, but we're not 100% sure this specific message is the one we'd want to change.

Based on the test case above, below are sample API payloads for a region with and without data. 

I noticed that some of the columns are missing entirely in the latter case- rather than "no data available", this specific error message roughly translates to "cannot reliably parse the API response as expected". Do you think this is worth investigating on your end before we develop a fix?

With data available in region:
http://www.type2diabetesgenetics.org/gene/getLocusZoom?chromosome=8&start=118105025&end=118265025&phenotype=T2D&dataset=GWAS_DIAGRAM_eu_mdv29&propertyName=P_VALUE&datatype=static

screen shot 2018-01-31 at 10 51 19 am

No data available in region:
http://www.type2diabetesgenetics.org/gene/getLocusZoom?chromosome=8&start=118081387&end=118171499&phenotype=T2D&dataset=GWAS_DIAGRAM_eu_mdv29&propertyName=P_VALUE&datatype=static
screen shot 2018-01-31 at 10 52 42 am

abought:
Hi, @benralexander ! Last week, you mentioned that an API bugfix had been shipped. I just tried the reproducible testcase above, and the payload still looks the same as in my previous comment.

Can you clarify whether the API bugfix is in production yet?
API URL: http://www.type2diabetesgenetics.org/gene/getLocusZoom?chromosome=8&start=118081387&end=118171499&phenotype=T2D&dataset=GWAS_DIAGRAM_eu_mdv29&propertyName=P_VALUE&datatype=static


Demo page: http://www.type2diabetesgenetics.org/variantInfo/variantInfo/8_118185025_G_A
```javascript
var thePlot = mpgSoftware.locusZoom.locusZoomPlot['#lz-47'];
thePlot.state.start = 118081387;
thePlot.state.end = 118171499;
thePlot.applyState();
```

Response:
```json
{"lastPage":null,"data":{"refAlleleFreq":[],"analysis":[]},"result_format":"flat","is_error":false,"header":"{}"}
```

benralexander:
The API code has been fixed on our dev machines, but the fix is not yet in
production.  We are hoping to deploy our newest code within the next couple
of days.

Thanks for you attention to these topics, Andy!

-Ben

benralexander:
Excellent.  Thanks, Andy!  I'll give it a try.

-Ben

abought:
Fix has been merged, and closing this ticket ahead of tomorrow's planned 0.7.2 release.

PR #112, opened 2017-09-19
Manhattan Data Layer

Frencil:
Work in progress, description forthcoming.


assets/js/app/Panel.js
```diff
@@ -798,6 +798,44 @@ LocusZoom.Panel.prototype.render = function(){
 };
 
 
+// Generate a chart-consumable ticks array from an arbitrary custom layout ticks definition.
+// Will pass through arrays but parse objects that invoke scale functions to generate ticks
+// from a data set (e.g. generating chromosome ticks for a GWAS manhattan plot).
+LocusZoom.Panel.prototype.generateCustomTicks = function(axis_layout){
```
abought: I haven't tried this with the pheWAS data layer just yet, but in principle I like this method better than my rough PheWAS approach, as it offers an option for the panel to customize tick display with defaults.  As far as I can tell, axis configuration is mostly the responsibility of the panel.

For PheWAS, our use case encompassed categories for colors as well as ticks. My current implementation relies on a new data_layer `category_field` option; for performance reasons I think it would be good to have the category data in a format that supported both use cases. (that way we don't need to iterate over all data so many times)  

I'd be happy to refactor my code as needed.

abought:
As suggested, I've merged in #110 (docs PR). Since this doesn't touch code, the merge conflicts should be fairly isolated.

Frencil:
Thanks @abought, the conflicts were pretty limited. This branch is getting a lot closer but still needs a bit more work. In the mean time though it is possible to see a functional proof of concept by checking out this branch and visiting the example page for it on localhost (something like [http://localhost/locuszoom/examples/gwas_manhattan.html](http://localhost/locuszoom/examples/gwas_manhattan.html), depending on where locally you serve the locuszoom library static website).

assets/js/app/Panel.js
```diff
@@ -971,10 +965,86 @@ LocusZoom.Panel.prototype.render = function(){
...
+        if (Array.isArray(layout.ticks)){
+            return layout.ticks;
+        }
+        // If the layout defines an object with a 'data' attribute then attempt to
+        // find a matching object on any of the child data layers. Use the first one
+        // found to drive the generation of a ticks array.
+        if (typeof layout.ticks === "object" && typeof layout.ticks.data === "string"){
+            var data = {};
+            this.data_layer_ids_by_z_index.forEach(function(data_layer_id){
+                data = this.data_layers[data_layer_id].data[layout.ticks.data] || data;
```
abought: I really like the idea of the automatic tick generation mechanism, and especially like the way this seems to allow customization from the external layout. Having the panel coordinate its own axis rendering makes a lot of sense.

One limit: in practice, I notice that most of our datasources return arrays (not objects)- even PheWAS data is actually an array. (although one use case is category-driven, there are other use cases- eg filter labels for highest pvalues- that rely entirely on the flat structure)

I think the solution is to decouple the duties of the data source from the internal data representation needed by any given datalayer. The plan is to add a new hook datalayer.getTicks (similar to the way that Panel.generateExtents / DataLayer.getAxisExtent works) that a parent panel will call as needed.

Hopefully will have enough to demo by later tonight- posting now in case the team has any thoughts on potential gotchas/ edge cases.

Frencil: @abought - An important distinction between panels and data layers... panels are the *masters* of axes. They own them and conceptually have all the logic for rendering them. Data layers may bind their data to an x and/or a y axis but that's it... and for good reason: a panel may have arbitrarily many data layers bound to the same axis.

The [Multiple Phenotypes (Layered)](http://statgen.github.io/locuszoom/examples/multiple_phenotypes_layered.html) example shows a panel with six data layers, along with a button-driven menu for fading/hiding/showing individual data layers as well as re-ordering them by z-index. As such, what happens when two or several datalayers, when asked to provide their ticks via the proposed `datalayer.getTicks()` method, provide different answers for the same axis?

I agree that most of our data sources come up short on this, but the idea of a data source is that it's custom middleware defined in the LocusZoom framework for doing arbitrary transformations on the data coming back from an endpoint/file so that it can be bound to attributes in a LocusZoom layout. Up until now tick generation, as driven by the layout, was done two ways:

1. Binding to a data field by name. Perfect for array-like data sets and continuous numerical fields.
2. Hard-coding an array of ticks into the layout. Bulky, and a stop-gap at best, but the only currently supported mechanism for categorical ticks.

The approach I have working currently on this branch introduces a third way without changing the current two:

3. Bind to a mapping object in data by name with a supported structure: keys as category IDs where each value provides a name (falling back to the key if not defined) and numerical values for placement on the axis.

True documentation would make this clearer. Regardless, the idea here is that the data sources `parseResponse()` method can be arbitrarily complex, and it's where we otherwise transform raw data into something we can bind to by other layout directives, so it's a good place to define this mapping object. On this branch you can see the GWASSource do that by following how the [chromosomes](https://github.com/statgen/locuszoom/pull/112/files#diff-a6139e2504656ffc08c847116c79ce29R882) object is built and also by looking at the data object on the manhattan data layer to see how it's organized (something like this):

```json
{
    chromosomes: {...},
    unbinned_variants: [...],
    variant_bins: [...]
}
```

I'll agree it's not great to have to loop through data layers looking for an attribute to bind to by the name in the layout. But then panels don't have data... only arbitrarily many child data layers that hold the data, and that's a *very* deeply rooted pattern.

I could see a way that the logic to create the mapping object could move to the data layer instead... that may be more what you're after with `datalayer.getTicks()`. The data layer needs access to the mapping object for its own rendering purposes (but again, it owns the data, so it has it whether it generates it or not as long as it's kept as a part of the data). But the logic to drive the creation of discrete ticks is probably best kept as close to the axis rendering those ticks as possible, and that's on the panel.

Here's one more angle to consider: both a PheWAS and GWAS/manhattan plot generally have at least *two* data layers... one for all the colorful scatter data, and one for the line of statistical significance. The latter can be just as opinionated about ticks, and being a simple line on a continuous number scale it would probably ask for a [pretty](https://github.com/statgen/locuszoom/blob/1c87f35a1aca9ed737a5081d8e806cf067bfe2ea/assets/js/app/LocusZoom.js#L181-L202) set of ticks that land within its extent, which it may or may not even be able to resolve without deferring to the other data layer(s). While it may not be able to say what ticks it would want discretely, it still needs an axis with a continuous numerical range in order to show up as a single line. So how it wants an axis (and therefore ticks) must be in lowest-common-denominator agreement with how the other data layer(s) want the same shared axis. That's easiest to achieve by expressing everything in terms of a continuous numerical range where categorical ticks map to discrete extents. If multiple data layers reported different things or incompatible things for ticks, how would the panel know what to draw?

abought: Thanks for the ideas- the point about complexity is a good one.

I just pushed some ideas that were in flight as you wrote this, to give a feel for the proposed mechanism. The implementation is based on enabling cooperativity while maintaining boundaries between layers (and keeping the final responsibility for the axis on the panel where it belongs).

In the [proposed change](https://github.com/statgen/locuszoom/pull/113/commits/78f98ae4ef388b0214a23d57b4e4e7c65f42e7cb), a datalayer has the option to *suggest* ticks, but the panel implementation makes the final call about how (or whether) to combine them. (and most layers won't ever suggest ticks at all) 

Specific use case: for something like category scatter, this lets us do things like guarantee that point colors (in the layer) would match axis tick colors (in the panel); I hadn't been able to find another way to do this short of having the layer overwrite parent layout properties directly. See screenshot from PheWeb sample case:
screen shot 2017-09-23 at 12 06 40 am


That said, pr 113 has been growing in scope as I've been delving into LZ, so think of this proposal as an experiment to see what is possible- I'm very flexible on this one. I've added a few notes to the PR description to cover some of the changes/ use cases.

Frencil: Just pushed up some changes that harmonize our two approaches somewhat. I do like the approach of letting data layers suggest ticks and the panel ultimately decide, and an array concat is a blunt but effective way to go about it in lieu of an actual use case of conflicting data layer ticks. I did, however, preserve the data binding approach I originally implemented and still have the data source doing the initial transform of the data, keeping the mapping object on the data layer's `data` object and identifying the binding in the data layer's axis layout by key name.

Will be back on sometime Sunday to continue progress.

abought: Thanks, Chris- I ran out of steam before the data part. :) if you think there's compatibility, I'd note we could define a single interface in PR113 and add features later. Remember, some version is slated to go out Monday and (while I consider design discussions with a good coworker to be a true joy) I also want to preserve your chance to enjoy your weekend.

Per data objects, long term this is one of the big challenges, I think. While at first glance it makes sense to let the data source control the custom shape of the data, I'm a little on the fence insofar as it limits interoperability across the system. E.g., what happens if it want phewas data sorted by category AND highest pvalue, as in labeled phewas plots? Do I need two requests to the same endpoint from separate datasources? Do I break features like labels.filters that assume they can call data.map?

That's basically the motivation behind dataLayer.getTicks- it lets the various plot layers communicate along well defined interfaces without having to rely on assumptions about internal states. 

(As an example from early drafts: if the fetcher here has to assume data is an object rather than an array, then inherently panel level functionality would seem to only work on a subset of layers).

abought:
Tiny side note: I noticed a few `lint` commits, so just thought I'd mention that for convenience you can run `npm test` when it's time to apply the full set of tests used by Travis.

I didn't add ESLint to the gulp task simply because some people find it distracting to worry about style guides during the normal development flow. The npm script should help though.

abought:
#112 merged! (watch for merge conflicts)

Thanks again for inspiring the better and more configurable tick generation mechanisms; even before being officially released, this PR has (hopefully!) had a positive impact.

The branch in its released state doesn't yet have good unit tests around the tick generation mechanism. Your branch has a good start there, and I'm totally ok having the remainder of what is needed delegated back to me once this branch (and the finalized mechanism) is ready.

abought:
Per meeting discussion today, this PR has been updated to point at the new `develop` branch, which is currently exactly the same as the contents of `master`. 

In the future, updates for ongoing development would pull new code from `develop` (and new PRs will target this branch). Apologies for any turbulence!

abought:
This PR is a bit stale; while we may reuse ideas in future I'm closing it due to merge conflicts.

PR #113, opened 2017-09-19
[rfr] Update PheWAS plot for new API

abought:
# Purpose
Expand the PheWAS plot example to work with the new API endpoint data sources, and to not rely on hardcoded tick mark labels.

Sample endpoint URL: `http://portaldev.sph.umich.edu/api_internal_dev/v1/statistic/phewas/?filter=variant eq '10:114758349_C/T'&format=objects&build=GRCh37`

This represents a new user-level feature to support T2D portal work, with a target external release date of Oct 1.

### PheWAS-related
- Expand `phewas_scatter.html` example plot with new variants
- Update the pheWAS datalayer to support a live API endpoint, and add JSON files of sample data for local development
- Add a new `category_scatter` plot type that can automatically generate tick marks and categories

### Internal hooks/ functionality
- Add a shared helper for generating subclasses
  - Allow extending known layers through the singleton registry
- Add an `htmlescape` helper for XSS prevention on user-provided data
- Add hooks that allow resetting panel and data layer state (for "plot interesting data" links that reuse existing panels). Demonstrate use on phewas scatter page
- (tentative proposal) Allow data layers to suggest ticks to the panel (but retain default behavior in most cases). This would allow a data layer to coordinate properties, eg for tick colors to match the color of datapoints in the layer. The panel would still be the final arbiter of display properties and could ignore the data layer ticks if desired.

screen shot 2017-09-25 at 12 08 37 pm
screen shot 2017-09-22 at 12 21 23 pm

# TODO/ Notes
- [x] Various TODOs inline (edge cases, data source configuration parameters, errors, styling, etc etc)
- [x] Evaluate migrating to more generic tick mark generation mechanisms in #112 
- Unit tests
 - [x] Subclassing mechanism
 - [x] htmlescape helper
  - [x] Extension mechanism for layers
  - ~Possibly test tick generation mechanism~

- [x] Discuss tick proposal, and make categoryscatter only return ticks for x axis (durgh)

Known issues:
- [x] Incorporate better sidebar variants (Ryan will choose good demos)
- [x] Plot breaks with unhelpful "missing ID field" error if there are no PheWAS variants in this region (eg if `payload.data = []`)
- [x] When applying filters, only the first value is getting labeled, not all that apply
- [x] if you click a point to "pin" the tooltip, and then jump to another region, it looks like the tooltip keeps getting redrawn (probably should be cleared when panel switches, maybe that's just an issue with the quick loadPheWAS() thing - *CONFIRMED* issue in scatter plot example before changes too


Frencil:
Tonight I've updated and solidified the tick generation pathways on #112. Here are some relevant parts of the diff:

* [Unified `generateTicks` method on Panel](https://github.com/statgen/locuszoom/blob/152fb7d5c06a88b587617d32b9d2f10c3c3ed850/assets/js/app/Panel.js#L968-L1043)
* [Unit tests for `generateTicks`](https://github.com/statgen/locuszoom/blob/152fb7d5c06a88b587617d32b9d2f10c3c3ed850/test/Panel.js#L264-L355) - These should demonstrate some basic approaches to all forms of tick generation, including binding to a data set, specifying a field within that data set, handling grouping and padding, etc.

welchr:
In some cases it looks like it's labeling a point with the trait label, and in other cases it isn't. Example: 

image

If you merge in my PR, the first example on the left sidebar will trigger the behavior, but the third link down will not.

abought:
> In some cases it looks like it's labeling a point with the trait label, and in other cases it isn't.

Yeah, this appears to be a regression. Labels are subject to filters (`p>=20` as default threshold), but it seems to be labeling only the first.

examples/phewas_scatter.html
```diff
@@ -171,14 +171,16 @@ <h3 style="color: #777">API (for developers)</h3>
...
 
-        var variantChrom = 10;
-        var variantPosition = 114758349;
-        var variantRefAlt = "C/T";
-        // Variants are expressed in format `10:114758349_C/T`;
-        var variant = [variantChrom, ":", variantPosition, "_", variantRefAlt].join("");
+        // Break the variant into constituent parts for setting plot state
+        var variantGroups = VARIANT_PATTERN.exec(variantForPlot);
+        var variantChrom = variantGroups[1];
+        var variantPosition = variantGroups[2];
```
welchr: When the page first loads it's not quite lined up on the right spot in the gene track, I think b/c variantPosition now ends up as a string at first. This seems to fix it: 

```
var variantPosition = parseInt(variantGroups[2]); 
```

abought: Update made; my fault entirely for not refreshing properly.

I'm going to squash the commit history tonight and consolidate what we will send out, and will do some basic regression testing first half of tomorrow with fresher eyes.

assets/js/app/DataLayers/scatter.js
```diff
@@ -466,3 +467,107 @@ LocusZoom.DataLayers.add("scatter", function(layout){
...
+        }
+        var categoryBounds = this.categories;
+        if (!categoryBounds  || !Object.keys(categoryBounds).length) {
+            return [];
+        }
+
+        return Object.keys(categoryBounds).map(function(category) {
+            var bounds = categoryBounds[category];
+            var diff = bounds[1] - bounds[0];
+            var center = bounds[0] + (diff !== 0 ? diff: bounds[0]) / 2;  // Center tick under one or many elements as appropriate
```
Frencil: Unless someone else had suggested PheWAS x ticks be centered on their extents and had a good reason to back it up I'd say this centering should be optional (driven by the axis layout) and default off. Compare what's showing in this branch currently vs. [existing LocusZoom PheWAS implementations](https://www.michigangenomics.org/genetic_data.html). It certainly makes sense sometimes to center a tick on its extent, but for PheWAS since the phenotype names can get long we went for the 45° rotation to help everything fit and with that it's a lot easier to track categories to their associated scatter points when the text anchor is aligned with the start of the extent.

assets/js/app/DataLayers/scatter.js
```diff
@@ -466,3 +467,107 @@ LocusZoom.DataLayers.add("scatter", function(layout){
...
+            return {
+                x: center,
+                text: category,
+                style: {
+                    "fill": "#393b79"  // TODO: Make colors dynamic and match category labels used for data
```
Frencil: If you assign an index argument at the start of that map function:

```javascript
return Object.keys(categoryBounds).map(function(category, index) {
```

You can then use the color array you set dynamically in the layout to set the tick fill here:

```javascript
    "fill": data_layer.layout.color.parameters.values[index] || "#393b79"
```

assets/js/app/Layouts.js
```diff
@@ -354,28 +354,30 @@ LocusZoom.Layouts.add("data_layer", "association_pvalues", {
...
-        field: "{{namespace}}x"
+        field: "{{namespace[phewas]}}x",  // Synthetic/derived field added by `category_scatter` layer
+        category_field: "{{namespace[phewas]}}trait_group"
     },
     y_axis: {
         axis: 1,
-        field: "{{namespace}}pval|neglog10",
+        field: "{{namespace[phewas]}}log_pvalue",
         floor: 0,
         upper_buffer: 0.1
```
Frencil: Minor, but the example feels a bit tight and upping this value (0.15 or 0.2) might give it a bit more breathing room.

assets/js/app/Panel.js
```diff
@@ -770,6 +770,58 @@ LocusZoom.Panel.prototype.generateExtents = function(){
...
+    // Parse an explicit 'ticks' attribute in the axis layout
+    if (this.layout.axes[axis].ticks){
+        var layout = this.layout.axes[axis];
+
+        var baseTickConfig = layout.ticks;
+        if (Array.isArray(baseTickConfig)){
+            // Array of specific ticks hard-coded into a panel will override any ticks that an individual layer might specify
+            return baseTickConfig;
+        }
+        //  TODO: Didn't have time to incorporate the broader ticks.data mechanism of PR 112 into this release; this method can be expanded in future
```
Frencil: FYI PR 112 is migrating away from the `ticks.data` approach to something a bit closer to what we see in this PR.

examples/phewas_scatter.html
```diff
@@ -116,33 +256,93 @@ <h6 style="float: right;"><a href="../index.html">&lt; return home</a></h6>
...
+        var phewas_layer = layout.panels[0].data_layers[1];
+        // Tell the layer to also fetch some special fields; otherwise the datasource will hide this info (TODO)
+        phewas_layer.fields.push("phewas:pmid", "phewas:description", "phewas:study");
+        phewas_layer.tooltip.html = [
+            "<strong>Trait:</strong> {{phewas:trait_label|htmlescape}}<br>",
+            "<strong>Trait Category:</strong> {{phewas:trait_group|htmlescape}}<br>",
+            "<strong>P-value:</strong> {{phewas:log_pvalue|logtoscinotation|htmlescape}}<br>",
+
+            "{{#if phewas:study}}",
+              "<br><strong>Study:</strong> {{phewas:study|htmlescape}}<br>",
```
Frencil: The extra line break leading this line creates extra unneeded white space in the tool tips and should probably be removed.

assets/js/app/DataLayers/scatter.js
abought: I've made the change suggested- good point about long category names. This is now configurable.

The panel can now pass special configuration parameters to `dataLayer.getTicks()` to control things that can't be overridden after the fact; category_scatter will recognize `position` values of `left|center|right`.

Colors are now shared between tick points and categories too. And as a final additional layer of reinforcement, the category name is part of the tooltips for each individual point, even in the default layout.

assets/js/app/DataLayers/scatter.js
abought: Phew, thanks for the reminder! This was an important TODO and I'm glad it made it in.

abought:
For the preview, I'm going to declare this branch complete as is and begin my own final round of QA + documentation work. Worlds of thanks to @pjvandehaar , @welchr , and @Frencil for feedback and helpful improvements.

If anyone has feedback before the end of the day let me know. Otherwise plan is to merge this by COB today.

This PR contains testing for a lot of the new stuff, but I'll aim to fill in remaining gaps in test coverage later in the week, as schedule permits. My sincere apologies for the messy commit history, but the rebase looked even gnarlier.

welchr:
Thanks to all of you for the hard work on this. It looks great. :) I don't think anyone here is a real commit history purist (other than maybe Peter...)

When I see it merge into master (or wait until you msg me the OK?) I'll send an email to the Broad folks. 

Though I do have one question on that. Once it's merged into master, could you forward gh-pages up to master so I can point them to http://statgen.github.io/locuszoom/examples/phewas_scatter.html (instead of the quick dummy portaldev page I was using before)?

abought:
Found one issue during my QA and fixed. I also did light regression testing of our other plot types.

Below are the release notes I put together, and I'll merge after my 2pm meeting. (AFAIK, only ChrisC has push access for the [npm package](https://www.npmjs.com/package/locuszoom), but that's in no way urgent)

LocusZoom 0.6.0 - Release notes

## User-facing features
- PheWAS plots are now flexible and customizable
- New "category scatter" layer type can automatically find and plot groups in plot categorical data (such as PheWAS)
- PheWAS data sources can now talk to a real API, and examples include documentation on what is required for a plot to work.
- Add a new field filter, `htmlescape`, which is helpful when using data from other people in your custom plots


## Internal improvements
- Most code now annotates JSDoc information about methods and input types. This internal documentation should help developers looking to extend and build on the core, at a level of detail beyond the traditional user-level documentation in our wiki. Feedback and improvements welcome!
- Better support for extending built-in functionality with custom code (new helpers for subclassing and `DataLayers.extend()`)
- Add helper method `plot.clearPanelData`, which is useful when reloading data on the same page. (arguments/api may change slightly in the future)
- DataLayers can now communicate with the panel to generate custom tick marks for specific data. (if `DataLayer` subclass implements `getTicks` method). As used in `category_scatter`, this provides a way to customize plots during the rendering process, without needing to hardcode tick marks or add extensive boilerplate code in every similar plot.

 
## Bugfixes
- Better error messages when API returns no data
- Fix issue with "plot interesting data" links, where plot behaved oddly when a different region was loaded (error messages and tooltips should now properly clear)


## Possible breaking changes
- Most methods now better enforce data types, to help avoid subtle bugs. This means that a string ("10") and a number (10) are no longer considered interchangeable. Please update your layouts.

welchr:
I'm not actually sure if the Broad uses our npm package. My impression was they used code from our repository directly. We should probably update it regardless just in case anyone else may be using it. 

@Frencil can you add Andy to the npm package so he can modify as needed?

Issue #106, opened 2017-06-14
Tooltips on annotation track do not show in "merged tracks" mode


welchr:
How to reproduce: 

1. Go to http://locuszoom.org/locuszoomjs.php
2. Add a dataset (I used DIAGRAMv3)
3. Add bed track (I used ENCODE / chromHMM / Chromatin State Segmentation by HMM from ENCODE/Broad)
4. Hover over segments of track - tooltip will show
5. Click "merge tracks" 
6. Hover over segments - tooltip will not show

Frencil:
Bug confirmed. Likely a regression (I swear tooltips worked on collapsed bed tracks initially...) from the other tooltip work that's happened recently. Will take a look.

Frencil:
Issue resolved in #116.

abought:
PR merged, closing issue. Thanks again!

PR #116, opened 2017-10-02
Fix Interval Tool Tip Behaviors

Frencil:
## The Problem(s)

As @welchr documented in #106 tool tips were not working for interval annotation tracks in merge-tracks mode *(Referred to in the solution as Issue A)*.

In addition, I noticed that while tool tips always worked in split-tracks mode, the bounding box to highlight the track wasn't appearing when a plot was first loaded in split-tracks mode but would appear after merging and re-splitting tracks *(Referred to in the solution as Issue B)*.

## This Solution

**Edited to reflect current approach**

This branch addresses both issues by refactoring the notion of how a data layer may have separate **status nodes** to class when applying a status (like highlighted or selected) *in addition* to classing the node most closely bound to the element by ID.

We see this on intervals as well as genes, where what we're rendering for data is not necessarily what will provide a visual cue that a status is applied (as opposed to a simpler scatter plot element which shows status with a CSS class change alone). Both of these data layers have "bounding box" nodes that render behind data and can even have a layout-defineable padding.

To abstract this behavior including showing status with a "shared" node (e.g. highlight the whole track for intervals but only when splitting tracks) this refactor introduces the concept of **status nodes** as described above and relies on a new method `DataLayer.getElementStatusNodeId()` to select the proper element to apply the status to. This method is abstracted for all data layers and will just return null, skipping any classing of status nodes that probably don't exist. But genes and intervals override this function:

* Genes merely appends "-statusnode" to the element's ID to get a statusnode ID (in a strict 1:1 relationship). This is essentially what it was doing before, but where it used to be repeated a few times throughout the data layer file now it happens in one place.
* Intervals now behaves just like genes when not in split tracks mode, but when split this function can return the status node ID for the "shared" node to show status (again: highlight or select) for the entire track.

These changes eliminate two undocumented layout directives: `hover_element` and `group_hover_elements_on_field` (the former appeared in genes and intervals and the latter only appeared in intervals). These were "dynamic" layout directives that wouldn't make sense for a user to define up front anyway, so it's arguably better that this can all be done more abstractly now. The main point of `group_hover_elements_on_field` was actually a way of saying "use the track split field as defined in the layout" but at the abstract data layer level... now the interval layer can express that in terms of the `getElementStatusNodeId()` override instead.

### DIFF
--- a/assets/js/app/DataLayer.js
+++ b/assets/js/app/DataLayer.js
@@ -163,6 +163,19 @@ LocusZoom.DataLayer.prototype.getElementId = function(element){
 };
 
 /**
+ * Fetch an ID that may bind a data element to a separate visual node for displaying status
+ * Examples of this might be seperate visual nodes to show select/highlight statuses, or
+ * even a common/shared node to show status across many elements in a set.
+ * Abstract method. It should be overridden by data layers that implement seperate status
+ * nodes specifically to the use case of the data layer type.
+ * @param {String|Object} element
+ * @returns {String|null}
+ */
+LocusZoom.DataLayer.prototype.getElementStatusNodeId = function(element){
+    return null;
+};
+
+/**
  * Returns a reference to the underlying data associated with a single visual element in the data layer, as
  *   referenced by the unique identifier for the element
 
@@ -724,13 +737,9 @@ LocusZoom.DataLayer.prototype.setElementStatus = function(status, element, toggl
     
     // Set/unset the proper status class on the appropriate DOM element(s)
     d3.select("#" + element_id).classed("lz-data_layer-" + this.layout.type + "-" + status, toggle);
-    if (this.layout.hover_element){
-        var hover_element_class = "lz-data_layer-" + this.layout.type + "-" + this.layout.hover_element + "-" + status;
-        var selector = d3.select("#" + element_id + "_" + this.layout.hover_element);
-        if (this.layout.group_hover_elements_on_field){
-            selector = this.group_hover_elements[element[this.layout.group_hover_elements_on_field]];
-        }
-        selector.classed(hover_element_class, toggle);
+    var element_status_node_id = this.getElementStatusNodeId(element);
+    if (element_status_node_id !== null){
+        d3.select("#" + element_status_node_id).classed("lz-data_layer-" + this.layout.type + "-statusnode-" + status, toggle);
     }
     
     // Track element ID in the proper status state array
--- a/assets/js/app/DataLayers/intervals.js
+++ b/assets/js/app/DataLayers/intervals.js
@@ -22,8 +22,6 @@ LocusZoom.DataLayers.add("intervals", function(layout){
         track_height: 15,
         track_vertical_spacing: 3,
         bounding_box_padding: 2,
-        hover_element: "bounding_box",
-        group_hover_elements_on_field: null,
         always_hide_legend: false,
         color: "#B8B8B8",
         fill_opacity: 1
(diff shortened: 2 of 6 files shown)


assets/js/app/DataLayers/intervals.js
```diff
@@ -186,8 +191,8 @@ LocusZoom.DataLayers.add("intervals", function(layout){
                     .data([interval], function(d){ return d[data_layer.layout.id_field] + "_bbox"; });
                 if (data_layer.layout.split_tracks){
                     bboxes.remove();
-                } else {                    
-                    bboxes.enter().append("rect")
+                } else {
+                    bboxes.enter().insert("rect", ":first-child")
```
abought: This was a good find! In #106 you refer to this as a likely regression; perhaps a unit/regression test could help prevent it from happening again?

Frencil: Ideally sure, more automated testing is generally a good thing. In practice, though, it would require automated testing against the composition of the SVG which is more of an integration/behavior level test, not a unit test. We could do it in mocha but it would likely be brittle.

Establishing a pattern for robust integration tests in LocusZoom is a would-be-nice that we may not have the resources for at the moment.

Frencil:
@abought see updates and new PR description. I've refactored things to be a bit more abstract as the behavior here was also partially present in the genes layer. This approach, while going beyond the original scope of the bug, provides a more general framework for data layers using a single shared element to show highlight or selection for a group of individual elements - something we may find ourselves needing to do again in the future.

welchr:
[Check interval track bugfix](https://trello.com/c/D3TAI2fz/1-check-interval-track-bugfix)

welchr:
Looks good to me now, thanks for the fix! 👍

abought:
> I've refactored things to be a bit more abstract as the behavior here was also partially present in the genes layer.

That is a remarkably robust answer to the question of whether it's a general mechanism; thank you!

I'll work on reviewing this now. The remaining wishlist item- better documenting layout behaviors- can be handled at any later time, independent of this being merged.

assets/js/app/DataLayer.js
```diff
@@ -724,13 +737,9 @@ LocusZoom.DataLayer.prototype.setElementStatus = function(status, element, toggl
...
-    if (this.layout.hover_element){
-        var hover_element_class = "lz-data_layer-" + this.layout.type + "-" + this.layout.hover_element + "-" + status;
-        var selector = d3.select("#" + element_id + "_" + this.layout.hover_element);
-        if (this.layout.group_hover_elements_on_field){
-            selector = this.group_hover_elements[element[this.layout.group_hover_elements_on_field]];
-        }
-        selector.classed(hover_element_class, toggle);
+    var element_status_node_id = this.getElementStatusNodeId(element);
+    if (element_status_node_id !== null){
+        d3.select("#" + element_status_node_id).classed("lz-data_layer-" + this.layout.type + "-statusnode-" + status, toggle);
```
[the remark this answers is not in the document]
Frencil: > Any thoughts on the best way to document this feature for developers?

The natural fit for this would be [in the documentation](https://github.com/statgen/locuszoom/wiki/Data-Layer#adding-custom-methods-and-functionality) where we've got the beginnings of a comprehensive section on authoring custom data layers. Really this section should be expanded to include the basic anatomy of a data layer subclass (including conventions for defining styles in general as well as what style patterns can map to automated behaviors), along with strategies for both developing a new data layer type from scratch as well as expanding an existing one.

I can set about adding to that documentation in general as I'm learning new angles on previously existing challenges in data layer creation with the work on the Manhattan data layer.

abought:
> Really this section should be expanded to include the basic anatomy of a data layer subclass (including conventions for defining styles in general as well as what style patterns can map to automated behaviors), along with strategies for both developing a new data layer type from scratch as well as expanding an existing one.

I agree about expanding- I'm not sure I'd be comfortable grouping both user and developer facing docs in the same section. Both are extremely valuable, but serve a different audience. (By proportion, most readers of that section will still likely be people who want to know what parameters to use in their layout JSON)

Glad to hear the manhattan layer is progressing! I will be turning some attention to credible sets display (and better color control customization) in the future; I'll let you know if I make major overlapping changes to docs on my end.

PR #117, opened 2017-10-23
Update analysis ID used in demo/index

welchr:
Database update caused analysis IDs to move around. This is the new ID
for the dataset previously referenced in the demo.


welchr:
Fixed the remainder of the IDs I could find. Note the `plot_builder.html` file doesn't work still - it's possible something changed in locuszoom's API a while back and no one really ever used the plot builder since then.

Frencil:
The plot builder began to grow stale when emphasis was shifted to building out the static site with examples right inside this repo. I've filed issue #118 to look at it and determine if it's even worth keeping, and if it is, modernize it.

Issue #39, opened 2016-03-08
Credible sets


MrFlick:
We had a request from Andrew Morris to see if its feasible to display the credible set of variants for a particular association peak.

This feature request would involve two parts
- [x]  A data source (or transformation) to do the calculation based on the p-values (or test statistics) of all SNPs in a region. Each SNP would get a dichotomous label: in or out of the credible set.
- [x]  A way to visualize the SNPs in the credible set. This could either be some transformation on the p-value points or a separate annotation track.

hyunminkang:
The calculation of credible set is relatively straightforward. There is a
NG paper where Andrew Morris is the first author, which would be one of the
easiest to implement.

Hyun.

welchr:
![](https://github.trello.services/images/mini-trello-icon.png) [Credible sets calculation](https://trello.com/c/UyKbQllE/2-credible-sets-calculation)

welchr:
We had some back and forth via email and wanted to update current status here. 

Goncalo/Hyun suggested we start with the simple statistic: 

![image](https://user-images.githubusercontent.com/165778/31912974-c35efd94-b813-11e7-9366-a4990a9d31f2.png)

Where Zi is the z-statistic for each variant in the region. 

Since we only have log p-values, we need to convert them back into a z-statistic (or directly to Z_i^2). Typically you would do this with: 

![image](https://user-images.githubusercontent.com/165778/31913338-eb53acb8-b814-11e7-9214-c275c13f5897.png)

Where phi^-1 is the quantile function for the normal distribution. The challenge is that some of our p-values are very small (smaller than float64), and we need to compute this in javascript (unless we push to the backend.) 

Hyun pointed out that for small p-values, Z_i^2 has a linear relationship with -log10 p-value. So the idea would be for log p-values below, say, 300 - we use a pure JS implementation of the normal quantile function. For larger log p-values, we use the linear approximation. 

That turns out to work quite well (example code [here](https://github.com/welchr/as241/blob/master/logp_to_z2.js)) except when we try to evaluate exp(Z_i^2). Unfortunately, it seems we can't calculate exp(Z_i^2) when Z_i^2 is > 709 with 64-bit floats. This corresponds to roughly a log p-value of 155. 

This is around where I stopped. I was planning to look into using logp in place of exp(z_i^2) or model it directly but got pulled onto other issues. 

Goncalo also suggested:

[...]

PR #120, opened 2017-12-11
[rfr] Credible set visualization options

abought:
🦑 This is a work in progress and should not yet be merged

# Ticket
https://github.com/statgen/locuszoom/issues/39

# Purpose
Implement credible sets with new visualizations. This PR exists to track state of work and make visible to others for collaboration. Some user-facing hooks are in flux or need unit tests added.

# Summary of changes
- Consolidate build tooling and add sourcemap support for easier debugging. Removes vendor files from repo because unminified are needed for sourcemap creation.
- Update node dev dependencies to newer versions (keep vendor build dependencies pinned_

- Add new "display options" dashboard component for flexible visualization choices (eg color by fields)
- New example page showing use of external credible sets library

# TODO
- [x] NPM release of credible sets library; point example page at a tagged release version (with better CDN link)
- [x] Better explanation/ intro docs within demo page (or link to docs for standalone library)
- [x] Unit tests for new layers/api hooks

## Known issues
- gulp-wrap does not support proper source maps so app build JS line numbers are consistently off by ~40 lines
- Pre-existing bug: two linked tracks can get out of sync if user pans around a lot: 1) drag a top plot (like association scatter) to the left of the screen about halfway across the range; 2) drag a bottom plot (like genes) halfway across the range to the right; 3) repeat a few times, and eventually the genes won't line up under the same place in the scatter plot.

# Testing notes
## New functionality
- New annotation track
- Dashboard button for display options. Should be able to control point size, point shape, color, and legend using existing layout directives.

## Potential regressions
- The build process has been reworked, and built assets are now deposited in the `dist` folder (to support direct NPM installs via Git URL)

# Release notes
LocusZoom 0.7.0
## New features
- Single-item annotation track: customizable track to identify points that meet certain criteria. Useful when annotating specific data points alongside other information like LD: allows rich display of many kinds of information on the same plot.
- Display options dashboard button: interactively apply different colors and styles to 
- New demonstration: annotate plots with region-based credible sets based on standalone `gwas-credible-sets` library

## Internal improvements
- Revamped build process; we now provide source maps for vendor files to facilitate debugging
- Built assets have been moved to consolidated `dist` directory


abought:
This branch is largely feature complete. Tomorrow will be on various tidying, testing, and user feedback, with a target to package and release by noon Friday at the latest.

This PR adds a new dashboard button + new data layer type. Whether before or after the holidays, any feedback on desired features (or implementation gotchas) would be welcomed. (eg @Frencil etc) The actual calcs are implemented in a [separate repo](https://github.com/statgen/gwas-credible-sets/).

Hope everyone has an excellent break!

files.js
```diff
@@ -66,12 +36,17 @@ var app_build = [
...
-// Vendor libraries. These are *only* libraries necessary for implementing the plugin.
-// Any vendor libraries needed solely for testing should not appear in this list.
-var vendor_build = [
-    "./assets/js/vendor/d3.min.js",
-    "./assets/js/vendor/q.min.js"
+// App, vendor, and helper files to be included at the top of each test suite
+// NOTE: Order of inclusion is important!
+var test_include = [
+    "./node_modules/should/should.js",
+    ...app_build,
```
Frencil: FYI this is the first appearance of ES6 in the repo and this changes the node dependency for development. I don't use node much so I was still on 4.2.6 (which is suitable for LocusZoom dev) but spread operator support wasn't added until 5.12.0.

Unfortunately upgrading node to current stable (9.2.1) broke other dependencies. Down(up?)grading to 6.4.0, the release that jumped from [59% to 95% ES6 support](http://node.green/) broke similar dependencies (sass-related) but in a slightly different way. Ultimately 5.12.0 worked.

So what I've learned:

* LocusZoom.js's development setup has historically targeted node **4.x**
* With this branch a **minimum node version of 5.12.0 is now required**
* Node **6.4.0 through current stable does not work** for our current development setup
* Old dev dependencies are a root cause of poor node support

I'd be curious to know what you're running locally, seems like we need to thread a needle on this. Those other packages are in big need of being updated.

Also, another dev note: I needed to manually delete existing built files (`locuszoom.app.min.js` in the root dir). Without that they got picked up and mapped by the new build process, barfing out a glut of lint errors.

test/unit/Layouts.js
```diff
@@ -1,30 +1,10 @@
...
-
-    // Load all javascript files
-    var src = [];
-    files.test_include.forEach(function(file){ src.push(fs.readFileSync(file)); });
-    jsdom({ src: src });
-
-    // Reset DOM after each test
-    afterEach(function(){
-        d3.select("body").selectAll("*").remove();
-    });
```
Frencil: Really nice to see this redundancy abstracted away, thanks for doing that!

package.json
```diff
@@ -12,24 +12,38 @@
...
   "author": "Christopher Clark <chriclrk@umich.edu>",
+  "contributors": [
+    "Matthew Flickinger",
+    "Ryan Welch",
+    "Peter VandeHaar",
+    "Andy Boughton <abought@umich.edu>"
+  ],
   "main": "locuszoom.app.js",
+  "engines": {
+    "node": ">=6.12.0"
```
Frencil: **BLOCKER**
As mentioned elsewhere I had dependency failures on node >= `6.4.0`, but `5.12.0` worked. Maybe pin this to 5.x until we can sort out dev dependencies in a later iteration?

files.js
abought: > I'd be curious to know what you're running locally, seems like we need to thread a needle on this. Those other packages are in big need of being updated.

Indeed! Coordinating which node version to use is always a bit of a headache.

I use a program called `nvm` to help me switch between different node versions (hopping between old and new projects is one reason I try to avoid global packages- there's no one answer that fits everything). The file `.nvmrc` gives a clue: I'm using node 8 LTS (carbon). I'm a little surprised that it didn't work for you- are you still having the problem after blowing away `node_modules` and reinstalling? (this PR does include dependency updates)

This also explains the change to the `engines` field of package.json: a warning to others to upgrade. A rough rule of thumb with node is to use odd-numbered versions; 4 is the oldest LTS, and 8 the newest. I'

The node version I use is captured in `.nvmrc`- so if you use

abought: > Also, another dev note: I needed to manually delete existing built files (locuszoom.app.min.js in the root dir). Without that they got picked up and mapped by the new build process, barfing out a glut of lint errors.

That's a fair point. My understanding is that few people develop directly on the code- is it fair to assume that we can tell them individually?

We previously git-ignored those files on `develop`, but not on `gh-pages`. 

In addition to being inconsistent, this meant that developers couldn't take easily take advantage of a semi-secret feature that lets NPM install a git repo from a specific commit hash (for people who wanted to drop the newest build into their project while still using package managers to track things exactly).

test/unit/Layouts.js
abought: ...I was trying to address intermittent test failures (due to slow jsdom start) and got carried away. ;)

Now if I could just get sourcemaps working (via getting `gulp-wrap-js` to behave or otherwise), we'd have a pretty convenient developer setup for debugging all around...

files.js
abought: Fixed typos in the above notes; they make more sense read on Github than from an email. Sorry about that.

The most common dependency issue will be sass, but the error should include a command to rebuild and fix the problem.

abought:
Added some unit tests, and at least for the exact method, results match R output. (haven't manually ported over all the extreme value cases)

@welchr, current goal is to package this up and cut an LZ release by ~12pm tomorrow, based on current state of branch. I'll hold off if you have change requests and/or if we can't resolve @Frencil 's tooling issue report by that time. 

Last call for tweaks!

abought:
Hmm. I'm tentatively marking the blocker as resolved, based on a few trials (below). If problems persist, please let me know and we can investigate further.

I've run the build/install process with 6.4, 5.12, 6.12 (LTS-6), and 8.9.3 (LTS-8), on old (6+ months) as well as new copies of the repo and on a different computer. I also tried with wiping node_modules after each install (to ensure bad dependencies weren't getting "grandfathered in").

The one issue I saw could be solved by good old `npm rebuild node-sass --force`; after that assets could be built normally on all machines/versions.

PR #123, opened 2018-02-01
Display results from burden and gene based tests

abought:
# Purpose
We would like to show burden test results in a LocusZoom plot. Burden tests can incorporate many masks, and be connected various ways of grouping variants- genes, intervals, etc.

Rather than a single visualization, we will start by rendering an HTML table- this gives us an opportunity to extend LocusZoom beyond the boundaries of the plot, so that it can be used as part of a broader collection of web page widgets that all update in response to the same data and events.

# Summary of changes
## New feature: "Subscribe to Data"
External widgets can now see the same data used by the plot, and update live as new data is received. For example, a table of GWAS variant data variants can update as the user pans left, or filter when a point is clicked. See `plot.subscribeToData(fields, successCallback, errorCallback)` for details.

## New feature: Extensions
The repository now contains a place to register functionality that may be useful, but is separate from (and not required for) the core LocusZoom experience. Each extension must be included separately into your site, after the core LocusZoom JS bundle.

### Dynamic URLs
The Dynamic URLs extension makes it possible to create easy sharing links to a plot. It provides a way to load the page at a specific view based on URL query params, and to update that URL as the user interacts with the plot. Clicking the back and forward buttons causes the plot to jump back to a previous view, enabling interactive experiences.

This works best with standard plots and common state parameters like `chrom`, `start`, and `end`. It is possible to serialize more complex behavior, but highly customized views (like dynamically adding panels) might not be a good fit for this extension.

## Events
Some items in this section represent **breaking API changes** and are marked accordingly.

New features
- Events can now be bubbled from panel to plot. This makes it easier to subscribe to events from a single source.
- Events now identify the place where they originated (eg `plot.genes` for the genes panel)
- Events have a new way of passing structured data to listeners.
- A new `element_selection` event has been added. This is unique from `element_clicked` in that it conveys whether a selection is being activated vs turned off.

**Breaking changes**
- When an event listener fires,`this` is now always the value of where the listener was attached. It is never used to pass data. (few events provide data; the change mostly affects usages of `element_clicked`)
  - Example: if you listen to a panel event at the plot level, `this` will refer to the plot instance. If you listen to the same event at the panel level, `this` will refer to the panel.
- All event listener functions now receive a single argument with a consistent structure: `eventContext`, with `{sourceID: string, data: *}`. Each event determines what, if any, `data` to provide. 
  - `sourceID` refers to where the event originated (eg panel) rather than where the listener was attached (eg plot level).

## The data requesting pipeline
There are times when multiple sources must be requested to build up a view. We have added some new functionality to add power to this behavior.

- The data chain now saves a copy of the raw response payload  (`chain.discrete`) that would be returned by a single source. This is useful when requesting information from several different sources that do not directly combine.
- Added a new type of `Connector` source. This is useful when a single piece of data (such as a calculation) is used by many kinds of data layers: the data can be requested once, with separate `Connectors` for each use case. 
  - For example, a calculation can be run once, and the results can be displayed as a table view (one connector) as well as matched to the data points on a data layer (another connector).
  - Previously the request and the merging of fields had to be part of the same data source

**Breaking changes**
DEPRECATES several data source methods: `parseData`, `parseArrayToObjects`, `parseObjectsToObjects`, `prepareData`

Provides stub implementations to avoid breaking legacy applications

`parseResponse` is now a wrapper for several discrete steps (`normalizeResponse`, `annotateData`, `extractFields`, `combineChainBody`). Custom data sources now have more control over how JSON data is parsed into records.

## Aggregation tests
- Add an example page based on raremetal.js, a library that performs aggregation tests in the browser.

## Misc bugfixes and improvements
- Bugfix: should now correctly render scientific notation with negative numbers

# TODO
- [x] Event mechanism/bubbling rework
  - [x] Allow layers to emit and bubble
- [x] Add url features to demo
- [x] Subscribe to data mechanism
- [x] Data chain revamp
- [x] DataSource connectors
- [x] Integrate burden test library, caching, and different assoc datasource
- [x] Many unit tests
- [x] Gather together doc update info


abought:
Demo feedback / action item checklist:

## Existing items, ticketed separately
- [ ] #137 Genes track 
- [ ] #138 Default dashboard button cleanup

## Demo improvement requirements
- [ ] Raremetal.js JSON definitions https://github.com/statgen/raremetal.js/pull/7
- [ ] Use a sample dataset that matches the sample JSON (and show matching sngle-variant pvalues on the top plot)
- [ ] Use the same gene name in the genes track (from genes source) and aggregation track (from the calculations / api)
- [ ] Adjust default # sigfigs in pvalue and score display
- [ ] Should all table columns stay (result / statistic)
- [ ] Improve prose in demo explaining the method and workflow

PR #128, opened 2018-04-12
Fix errors where plot region has no data

abought:
⚠️ Replaces #111 by rendering the error message irrelevant

# Purpose
When an association plot is scrolled to a region with no data, cryptic error messages appear. (even after fixing the portal API format) Instead, we would prefer a behavior in which that data layer can render with no data points, and without preventing the rendering of other layers.

Before
screen shot 2018-04-12 at 2 09 06 pm

After
screen shot 2018-04-13 at 11 30 30 am

After fixing the portal API response, it was discovered that the LZLD datasource had a hard-coded assumption of at least one association record.

# Summary of changes
Introduces a new concept of "dependent" data sources. This covers sources like LDLZ, which are designed to annotate records fetched from a previous step in the chain. These sources have a strong coupling relationship to another type of data.

Since the LD source only adds meaning in the context of other data: if a prior source (like association) returns no data, dependent sources should skip their own request step entirely.

## TODO
- [x] Fix a rendering issue where the "line of significance" data layer is drawn diagonally in the absence of other data . (y-axis scale issues)

## Other
The build process now supports creating accurate source maps for `locuszoom.app.min.js`, to facilitate debugging.

# How to test this
This error scenario can be replicated using "offline mode" of the provided index.html example, and using the following as the contents of `staticdata/assoc_10_114550452-115067678.json`:

```json
{
  "data": {
    "analysis": [],
    "variant": [],
    "chr": [],
    "position": [],
    "refAlleleFreq": [],
    "ref_allele": [],
    "pvalue": [],
    "log_pvalue": [],
    "scoreTestStat": []

  },
  "lastPage": null
}
```


abought:
Ben has reported success with these changes. I'll add one more unit test in #123 (since that branch adds sinon support, useful for stubbing data sources).

PR #130, opened 2018-04-18
Phewas override colors

pjvandehaar:


abought:
Thanks for the PR! I'll try to gin up some sample data tomorrow and try this out, but otherwise looks good.

I'll also try to fix the merge conflicts that crept in from #128 . In the future, I'll always be regenerating the built assets during a release- we can reduce merge conflicts by not checking in the dist files on every PR.

abought:
Looks good! I added a few minor tweaks ( 4de2c03 ) based on initial testing and will merge now.

PR #136, opened 2018-05-24
[rfr] Allow forest plots to operate on dynamic data from the server

abought:
## Purpose

The existing forest plot demo had nice features, but required specifying the labels and layout in advance.

This PR adds a new `category_forest` mechanism that allows labels and extents to be set based on the data from the server (similar to the `category_scatter` used by our other PheWAS option).

It also upgrades the examples to use features characteristic of real sources (esp field namespacing). This should help reduce the example barrier to entry when using live server data. (since StaticJSON bypasses some common source functionality).

## TODO
- Documentation
- Open for user feedback from users currently investigating our forest plot options (@benralexander , @pjvandehaar  )
- Tentative release goal 0.8.0, but could be moved up independent of burden test additions if needed.


abought:
Removing 0.8.0 milestone. 

I haven't heard back from testers yet, so we'll push this to a followup release, and perhaps explore a more realistic demo using live server data.

Issue #124, opened 2018-02-26
Annotate GWAS Catalog SNPs in LZ plot


sgagliano:
Have an option to highlight variants that appear in the GWAS Catalog in the LZ plot, and if possible also have a way (maybe a toggle box) to see with which trait(s) that variant is associated with in the GWAS Catalog. 
Purpose of this feature: it would provide a quick visual way to identify whether the locus is known to be associated with a trait of interest, even if the reference variant itself is not known to be associated.

abought:
I could see a use case for this. The API docs [suggest](https://portaldev.sph.umich.edu/docs/api/v1/#overview-of-api-endpoints) that we do have rsID information in our database, though I can't seem to find it in the response payloads for any given SNP...

That said- our API server isn't the only way to get data into LZjs- you can use your own custom datasource if you have your own project using LZjs. 

I think the basic tooltip link and maybe some visual convenience would be feasible to add by using existing features- let me know how I could help?

welchr:
We would need to implement the GWAS catalog API and database tables, which is a small amount of effort since the data is so small. 

If I recall correctly, the reason we needed to mirror this was because the EBI GWAS catalog API did not support region queries, which we need for LZ. 

Sarah notes they would like to have this for their paper, but sounds to be up in the air on when they would like to submit. This of course has to take a back seat to the portal priorities.

abecasis:
Hey Ryan,

I really would love to see these annotations enabled and I would be happy
to bump them above some of the portal priorities. :)

I am not sure if we should implement an API direct to the GWAS catalog,
since their version of the data often has a lot of junk mixed in. It may be
easier for us to create our own hits table and that would let us (for
example) annotate UK Biobank peaks or some other interesting results of our
choosing.

Goncalo

abought:
Spoke to Sarah a bit more, and it seems there are two approaches.

The "easy" version of this feature would be for our API to return rsID alongside each variant's other data (chrom/pos/ref/alt), or "null" if the variant had no rsID. This would answer the question "are these SNPs known in the catalog at all" and would be enough for us mark "catalog variants" in the plot & put links in tooltips.

The more labor-intensive version would be to link the catalog to specific traits ("show me SNPS in this region that have been associated with Alzheimer's"). 

It's possible this could be done in two stages, depending on just how much a complexity jump is involved for the second version.

abecasis:
Can we wireframe this? We might be thinking about different things. This is
what I imagine:

[...]

PR #143, opened 2018-09-14
[rfr] Source + UI to mark known GWAS catalog SNPs

abought:
# Purpose
Per #124 , we would like to display GWAS catalog SNPs in a LocusZoom plot. This has utility beyond pheweb.

This PR adds a datasource that exposes the Portaldev api endpoint. See companion ticket for feature discussion. The UI side will be defined once we decide what filters and fields to apply.

screen shot 2018-09-21 at 2 46 24 pm


abought:
Merging this iteration into the `develop` branch so we can start using with PheWeb, but I'm very open to further improvements on request. My current hope would be a LZ 0.9.0 release ~next Wednesday (9/26).

abecasis:
You are too fast.

I mean the following (as an example):

Go to
https://abought.github.io/locuszoom/examples/gwas_catalog.html?chrom=10&start=114458349&end=115058349
(the TCF7L2 region in your examples).

There is an association to Glycated Hemoglobin Levels.

The association is sourced from a paper titled "Impact of common genetic
determinants of Hemoglobin A1c on type 2 diabetes risk and diagnosis in
ancestrally diverse populations: A transethnic genome-wide meta-analysis."

Can we link there? The link would be:

https://www.ncbi.nlm.nih.gov/pubmed/28898252

I think the GWAS catalog entries include the PMID (28898252 in this case).

G

Issue #139, opened 2018-08-10
[request] New loop track datalayer


abought:
We have discussed several potential visualization types recently that might be useful in LZ.js. This ticket exists to track notes on the current state of ideas.

- Filled curve (sample use case: ATAC-seq). See [T2D Portal implementation](https://github.com/broadinstitute/dig-diabetes-portal/blob/c2b8dcffa7977b42e072e8efdf4d0850e5e5b355/web-app/js/lib/dport/locusZoomPlot.js#L298) for a sample of something we could adapt to core.
- Gene "loop track" (sample use case: HiC)

abought:
Both filled curve and loop track have since been implemented; closing this ticket.

PR #147, opened 2018-10-03
Improve handling of filled curves

abought:
References #139

## Purpose
Improves rendering of filled curves when using `style.fill` config option on the existing line layer.

The previous option worked by implicitly closing the path (joining first and last point). This produced undesirable results if the endpoints were not already flat with the y-axis (see example). 

## Summary of changes
The new behavior changes the mechanism by which the `path` is generated to more explicitly fill the area between the data and the y-axis. 

Before:
screen shot 2018-10-03 at 4 20 28 pm

After:
screen shot 2018-10-03 at 4 35 13 pm

### DIFF
--- a/assets/js/app/DataLayers/line.js
+++ b/assets/js/app/DataLayers/line.js
@@ -2,7 +2,7 @@
 
 /*********************
  * Line Data Layer
- * Implements a standard line plot
+ * Implements a standard line plot, representing either a trace or a filled curve.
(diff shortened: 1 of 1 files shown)


abought:
Implementation very similar to one used by the T2D portal.

Will show this to a few prospective users. If there are no further requests, I'll merge in ~1wk.

Issue #144, opened 2018-09-14
[request] Forest plot improvements


abought:
Requested by @benralexander, as potential followup items to #136 . Ticketing to track for later.

# Purpose
We would like to improve the scaling capabilities of forest plots. In particular, we would like to provide users with a way to handle very wide confidence intervals, even when a few such points could swamp the display of most other data.

## Notes and sample data
> It's going to be the large beta values ( positive or negative) that are often the most interesting.  If those are precisely the values that users can't see them they will find the tool to be very frustrating.

> ...have the zoom capability work for the mouse.  That way you could expand and contract the X axis to whatever range you like.
> This would be parallel to the way that the standard association plot works -- it zooms only on the X axis, and leaves the y-axis constant.

He references a sample test case to be `rs13266634` "which I think I remember as being 1.4.  The confidence interval goes from some negative number to maybe 3."

# Options
Two approaches come to mind, together or individually.

1. Implement scaling/ zooming as suggested
2. Add a tooltip for data points that provides the exact beta, pvalue, and confidence interval limits associated with each mark.

abought:
Additional requests based on a user feedback meeting (courtesy of Ben):

1. Sort by effect size, would like to be able to set p-value cutoff
2. Is there a way to use the size of the box to denote relative sample size? Commonly done for forest plots, so should be intuitive for users. Not sure if binary vs continuous traits are comparable in the same plot. (user says is ok if take natural log of odds ratio)
3. Default view should be to show the most significant associations at the top

abought:
Additional requests:
- Allow forest plot to be filtered, eg, significant traits only
- Control sort order
- Control confidence interval range to help focus on standard errors of the most significant traits, where error bars might be more narrow.
- Frequent question: is data filtered, or not present? Provide a visual indicator
- Would be nice to have visual connection between phewas and forest plots- similar coloring, threshold for significance

abought:
The broad has an alternate implementation of forest plots. In LZjs 0.13, we'll move the forest plots to an extension, as it is not a highly used feature.

PR #148, opened 2018-10-09
[rfr] Forest plot improvements

abought:
References: #144 

# Purpose
Further improvements to forest plot display based on portal usage. (h/t @benralexander)

# Summary of changes
- Autoscale `category_forest` x axis to data (including confidence intervals). Add new `_getDataExtent` hook for the case where axis extent depends on more than one field.
- Fix tooltip highlight/selection bugs in forest plots (related to generated element IDs)
- Restore broken tooltips to forest plot example layout
- Add show/hide legend button to forest plot example layout

# TODO
- [ ] Investigate a mouse wheel zoom feature (currently very buggy in this plot type)

### DIFF
--- a/assets/js/app/Data.js
+++ b/assets/js/app/Data.js
@@ -231,7 +231,7 @@ LocusZoom.Data.Requester = function(sources) {
         });
         //assume the fields are requested in dependent order
         //TODO: better manage dependencies
-        var ret = Q.when({header:{}, body:{}, discrete: {}});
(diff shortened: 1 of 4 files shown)


abought:
Scrolling support update: on the forest plot, the following `panel` level `configuration options work:
```
interactions: {
    drag_background_to_pan: true,
    drag_x_ticks_to_scale: true,
}
```

However, the following additional desired interaction behavior is giving errors, and will take more digging to implement correctly.
`scroll_to_zoom: true,`

PR #150, opened 2018-10-15
Annotation track improvements

abought:
References #124 and https://github.com/statgen/pheweb/compare/8cdb7...pull-112-head#diff-007e08bf75bfe53b85b14034803e369c

Expand suggestions by @pjvandehaar and incorporate into core codebase:
- Make it easier to click on 1px-wide annotation lines
 - Configurable `tooltip_positioning` option for annotation track. Supports two fixed positions (`top` and `bottom`) and one dynamic option (`middle`).

## How to test this
The effect of each position option can very quickly be tried in the console: 

`plot.panels["annotationcatalog"].data_layers["annotation_catalog"].layout.tooltip_positioning = 'bottom'`

## Alternatives considered
- Always position above / below: each layout has different needs.
  - The default option ("middle") makes sure that the base layout is reusable without customization. 
- Horizontal tooltips: since all the spread of data is horizontal, this was too likely to overlap with data

### DIFF
--- a/assets/js/app/DataLayers/annotation_track.js
+++ b/assets/js/app/DataLayers/annotation_track.js
@@ -16,7 +16,9 @@ LocusZoom.DataLayers.add("annotation_track", function(layout) {
     // In the future we may add additional options for controlling marker size/ shape, based on user feedback
     this.DefaultLayout = {
         color: "#000000",
-        filters: []
+        filters: [],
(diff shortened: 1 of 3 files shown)


pjvandehaar:
The separate `<g>`s aren’t needed. Perhaps the code would be clearer without them.

PR #161, opened 2019-06-20
Move to native Promise.

lrowe:
Avoids extra dependency on Q.

### DIFF
--- a/assets/js/app/LocusZoom.js
+++ b/assets/js/app/LocusZoom.js
@@ -245,42 +245,42 @@ LocusZoom.prettyTicks = function(range, clip_range, target_tick_count) {
  * @returns {Promise}
  */
 LocusZoom.createCORSPromise = function (method, url, body, headers, timeout) {
-    var response = Q.defer();
-    var xhr = new XMLHttpRequest();
-    if ('withCredentials' in xhr) {
-        // Check if the XMLHttpRequest object has a "withCredentials" property.
-        // "withCredentials" only exists on XMLHTTPRequest2 objects.
-        xhr.open(method, url, true);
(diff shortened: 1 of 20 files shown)


assets/js/app/LocusZoom.js
```diff
@@ -245,42 +245,42 @@ LocusZoom.prettyTicks = function(range, clip_range, target_tick_count) {
...
-        xhr = null;
-    }
-    if (xhr) {
-        xhr.onreadystatechange = function() {
-            if (xhr.readyState === 4) {
-                if (xhr.status === 200 || xhr.status === 0 ) {
-                    response.resolve(xhr.response);
-                } else {
-                    response.reject('HTTP ' + xhr.status + ' for ' + url);
+    return new Promise(function (resolve, reject) {
```
abought: Reminder to self: add a deprecation warning to this function. We'll preserve it for custom deployments, but most new usages can probably be served by the built in `fetch`.

abought:
Thanks for the submission! Before I merge, I'd like to do a round of regression testing (probably early next week) on various deployments- but this is something I've been wanting to see deprecated for a while.

We always welcome pull requests- let us know if you have any other ideas in mind. `Promise` is ripe for use because it is very amenable to polyfilling, and we have recently been working on better support for modern build workflows that can do that automatically. Much of the modernization work is being piloted in our LocalZoom  tool, eg [module support](https://github.com/abought/locuszoom-tabix/blob/develop/src/util/lz-helpers.js#L1).

abought:
Regression tested in FF, Chrome, and Safari with no problems observed. A thorough and well executed conversion- thanks for your contribution! It will go out in our next release.

PR #181, opened 2020-04-28
locuszoom co-accesibility track basic

Parul-Kudtarkar:
LocusZoom Track to display links between genomic sites such as chromatin loops (from Hi-C/CHi-C) or co-accessibility (from single cell data).  The tracks containing loops(polyline triangles) between two linked loci and where the loop height is based on a value (e.g. the significance of a Hi-C loop).


abought:
I've incorporated this PR into #182, and will continue extending the demo with new features as we go. Closing this PR to avoid duplication, but we are very much looking forward to seeing this work folded into LZ! 

For any further additions, I can coordinate branch access if needed just so that all work is in one place.

As a fun aside, the main repo is at ~999 commits. I think it would be fitting for our 1,000th commit to come from a collaborator!

PR #182, opened 2020-05-06
New arc/coaccessibility track

abought:
Builds on #181, by @Parul-Kudtarkar .

# Purpose
New "arcs" type DataLayer, for visualizing co-accessibility (and possibly other datasets) in a local region track.

# Summary of changes
- New `arcs` layer type draws a symmetric arc between a start and end point (x0, 0)... (x1, 0), with a specified height for the arc. The track accepts filters to limit what data is shown, and each loop can be colored independently. (line weight, stroke width, and opacity are configurable for all loops at the layer level)
- Scaling functions now accept an additional argument, `index` in the data array. This allows scaling functions to treat elements differently based on ordering. For example, a coloring function can cycle through a set of options to draw each point in a different color.
- [ ] New coaccessibility demo that incorporates some advanced LZ features (arc track, add additional tracks to plot, click to match points...)

# TODO
- [x] Unit tests for scaling functions and layers
- [x] New LZ extension, TabixSource, allows reading from remote Tabix-indexed files in a semi generic manner. An example is provided for BED track data.
- [x] Rework the internal mechanisms for tooltip generation to enable greater reuse. (in some layers, including base datalayer, the default tooltip generation mechanism does not work)


Parul-Kudtarkar:
The updates look pretty cool! Thanks @abought !

abought:
There are a few improvements we'd like to make (eg more plot fields), but the basic demo works and is in a good place.

Merging now so that I can combine this with other refactoring and code cleanup in progress for the intervals track. Happy to keep making improvements!

abought:
Incidentally, Parul's contribution just happens to be our 1,000th commit to the LocusZoom.js repo. Thank you!

Parul-Kudtarkar:
This is great! Thank you for the development and collaboration.

Issue #196, opened 2020-07-02
Provide mechanisms for custom filtering of data layers


abought:
Requested by @massung , and targeted for v0.13 (two releases hence) based on current work queue. (EDIT: moved to 0.12)

There are several use cases for which people would like to filter the data elements that are displayed on the plot.

We would like:

1. New UI to expose our existing filter syntax via toolbar buttons
2. Implementation to accept that UI across several data layers. ("arcs with score above significance threshold", "significant assoc variants", "specific intervals)
3. A low-level method `BaseDataLayer.setFilter(afunction)` , to manually declare an arbitrary filter function that does not fit within the configuration mechanism.

There should be an order of precedence for each of these options. @Parul-Kudtarkar and @psnehal have also suggested such a feature recently and may have interesting suggestions.

kbruskiewicz:
Importantly the filter should take effect on UI refresh, or otherwise setting the filter refreshes the UI using filtered data.

massung:
I also wanted to note that I think it's 100% reasonable for LZ to assume the filter [function] is dealing with 100% immutable data. For example, I think this would be an unreasonable filter function (using Vue as an example, but would be true of any reactive framework):

```js
let pThreshold = this.$store.state.pValueThreshold;

lz.setFilter(item => item.p_value < pThreashold.get);
```

I think the above _could_ be supported if a watcher (on `$store.state.pValueThreshold`) called `lz.refresh()` or similar, which forced a redraw. But, there would be no way for LZ to just know the value changed on its own. I would expect that when the page state changes, we could (and maybe should) be expected to provide LZ with a new filter function.

abought:
Work is underway in an open PR that describes the syntax: https://github.com/statgen/locuszoom/pull/198

One quirk is that some behavior may not be well defined for certain data layer types; if there is not a clear use case for eg genes track, then I'm not sure it's worth the investment of time and energy internally to de-tangle the edge cases. It makes most sense, and helps deliver faster, if we can narrow down specific visualizations/use cases for this feature.

Based on the provided example, I assume that @massung is looking to filter, eg arc, annotation, and scatter plot? Two syntaxes (declarative and imperative) are provided in the proposed PR; tomorrow I'll work on UI.

Agreed with Jeff: given how we implement separation of concerns, the plot should not assume that data is changed outside of `getData`. The UI widget will work via layout mutation (modify `data_layer.layout.filters` and re-draw).

massung:
[...]

PR #198, opened 2020-07-22
Allow data layers to render a filtered set of points

abought:
Ticket: #196 

## Purpose
Implement a way for a data layer to "filter" its data and show only a subset of points. Two syntaxes are supported:
- Declarative, with built-in filtering operators (`layout.filters = [ {field, operator, value}... ]`). May be defined before initial render.
- Imperative, by defining an arbitrary filter function on the layer after the initial render is complete (`layer.setFilter(func)`). The user-defined function is a callable with the same arguments as `array.filter`. It is applied on each render.
  - Layouts must be JSON serializable; there is not a mechanism to define imperative functions before first render.

## Summary of changes
- Remove unused filtering methods (addField, setElementStatusByFilters)
- Unify filtering syntax to be the same across labels and top level layouts. **Breaking change**: all now use `{field, operator, value}` (filters no longer declared as arrays)
- Add data-filtering syntax to scatter plot, genes, forest, etc. This will require further testing, as for some datasets the behavior is not very well defined.

## TODO
- [x] Implement basic UI toolbar widget that allows quick filtering based on user-provided values
- [x] Add unit tests
- [x] Define behavior when user enters a filter, then clears value (esp for exact match)
- [x] Test on all layer types that support filtering functionality


abought:
Screenshot of user-selectable filter functionality (in progress):

Jul-23-2020 16-57-09

abought:
Merging to facilitate testing, but open to further revisions before 0.12 final.

Issue #215, opened 2020-11-10
Allow template string syntax to support "else" branch on if statements


abought:
Sometimes in tooltips, we want to control what is rendered based on existing data. Eg, a tooltip might contain an action button that says "show label", and once clicked the text should change to "hide label".

Currently our template syntax supports only if statements ("if a value is present at all"). An else block would be helpful for this use case.

Currently the if statement only checks if a value is present. Some additional development would be necessary to assess whether the current grammar allows smarter conditionals:
* "if value === x"
* else if

PR #217, opened 2020-11-17
Expanded and revised conditional analysis feature

abought:
## Purpose
We would like to add a way to perform conditional analysis from within locuszoom.

Although an old demonstration existed, it was never fully implemented, and it revealed a number of design limitations in how LZ handles things like annotations. We should unify the demo with new mechanisms, implement calculations, and add several features.

## TODO:
- Add a display options button to toggle to display of results
- Fix various UI and alignment issues
- Lock scroll once model building starts
- Add a "suggest best" option that adds the LD ref var to the covariates list.
- Add a UI button to add user-specified variant, or search by various mechanisms.

## Summary of changes
- New `element_annotation` event allows external widgets to respond to annotations (user selected lists of significant points that are remembered across re-render: eg "use this element in a calculation and redraw it with a custom shape")
- [ ] Consider a "custom event" syntax that allows any event to be sent from custom widgets (no longer limit to pre-set event hooks). Example: "calculation complete" can be used to communicate with a data table, and show results only when some defined step in a UI flow is complete.
- [ ] Improve template syntax to support toggles like "add/remove" (#215)
- [ ] Implement conditional analysis calcs in raremetal.js
- [ ] Implement an LZ data source that uses the selected items to drive the calculation

### DIFF
--- a/esm/components/data_layer/base.js
+++ b/esm/components/data_layer/base.js
@@ -212,6 +212,7 @@ class BaseDataLayer {
             this.layer_state.extra_fields[id] = {};
         }
         this.layer_state.extra_fields[id][key] = value;
+        this.parent.emit('element_annotation', { element: element, field: key, value }, true);
         return this;
     }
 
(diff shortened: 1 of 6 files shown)


abought:
On hold until we better define the UI and impact requirements with the T2D portal.

Issue #186, opened 2020-06-01
Documentation improvements


abought:
# Purpose
The LocusZoom documentation is incomplete, and manually maintained. This makes it difficult to keep up to date.

# Scope of work
## Changes to existing documentation
- [x] Copy existing documentation from wiki to GitHub repo, as a historical reference for older versions. Blocker for 0.11 release (at least one release must have time-stamped versioned docs before we rewrite internals in 0.12!)
- [x] Convert all doc information to JSDoc format comments
- [x] Transition doc builds to JSDoc and add them to github pages.
- [x] New topic guides, and create a section on the wiki to cover this.
- [x] Add/expand key information

## New topic guides
Aimed at helping developers to get started with complex topics in LZ, for which documentation and examples may be especially dense or scarce.

- [x] Working with data
- [x] Layouts and rendering (revise existing tutorial, cover pieces and key options like scalable parameters)
- [x] Guide to interactivity
-  ~Future: extending LocusZoom with plugins: creating custom data layers, sources, etc.~
- [x] Harmonize JSDoc with wiki
 - [ ] Remove content of the wiki as of v0.13.0 release
- [x] Make doc structure consistent, add landing pages

abought:
We've added many mechanisms for interactivity in recent releases, and these are presently under-documented. (semi-public APIs that are being refined based on usage)

Modifying the plot programmatically is a complex topic that requires knowledge of the internals, and it would be a good candidate for a guide.

PR #231, opened 2021-01-20
Port all documentation to JSDoc

abought:
Refs #186

The current documentation is a wiki, which is not tied to a specific version, and is updated by hand. It omits many layout options and is cumbersome to keep accurate/up to date. Some of the wiki info reflects features that were never fully implemented or removed, confusing users mightily.

We have been building JSDoc comments into the codebase to prepare for a transition to machine-generated API docs. Build, test, and expose those docs to replace the wiki, with improved information about layout options.

New things to document:
- Event hooks (and signatures)
- Reorganize guides and docs into a consistent structure
- Check public and private api specs
- Capture missing layout directive options (esp adapters, widgets, and layers)
  - Can we add markdown to the adapters section to explain the data chain to developers?
- Template adjustments (eg remove build date from footer)
- Provide header info and markdown sections in module docs explaining how to load an extension.
- Cross link with tutorials, to show idiomatic usage and encourage users not to interact with classes directly.
- Improve layout guide with more details for using scalable layout directives


abought:
Core updates complete. Will merge this in order to combine with other guides in progress- it won't go live until the "working with data", "interactivity", and "rendering with layouts" tutorials are completed. (a trio of key concepts in a user-facing style, so they can get help without having to know about LZ internals first)

PR #269, opened 2022-05-02
doc: Fix "separation of concern" list display

sir4ur0n:
Add a missing empty line before the beginning of the list

See current display at https://statgen.github.io/locuszoom/docs/guides/index.html

image


abought:
Thank you for catching this!

PR #277, opened 2022-07-04
Add enrichment example, and add link to each example source code

sir4ur0n:
This PR could technically be 2 smaller PRs, but I chose to put them in a single PR as they are rather small. Let me know if you would rather I split those.

I recommend reviewing commit by commit.

* I noticed there was an `interval_enrichment.html` example page, but it's not referenced from the "Examples" page, so I added it. I also added a preview screenshot, respecting the 250x125 format of others
* I have several times felt the need from an example page to view the actual source code (not the end-result HTML code, because it contains the LZ plot), so I suggest adding a link to each example source code. Moreover, as some examples are in the `ext` subfolder, it can be even a bit confusing to readers without the link. I think it also slightly helps readers understand that each example is self-contained as we link to a single file.

What do you think?

image

image


abought:
I love the link to source code! Because of all the configuration options, examples are an important part of documentation. Making the code easier to find is an excellent idea.

The enrichment demo is sort of a weird case, and hiding it was deliberate. Without going into the details, it was delivered at the request of a project partner, who then realized that the design they asked for didn't match the data they were producing. We kept the demo there (but separate and hidden) until a potential use case could be found.

sir4ur0n:
I just force-pushed a new version: basically everything is the same except it's no longer in `index.html`.

I kept the screenshot for consistency and in case there is a use case in the future, but if you prefer I can remove it (or you can remove it before merging, as you wish)

abought:
Thanks! Looks good.

After merge, I will make a tiny change so that the links point to `master` instead of `develop`. That way, docs will always point to the latest release of software, and links will not break during in-progress refactors.

PR #282, opened 2022-09-23
Fix wrong type of BaseDataLayer.layout.[x|y]_axis.field

sir4ur0n:
The types were incorrectly flagged as `number` instead of `string`


abought:
Thanks for catching this!

